PR #143, opened 2018-12-19
WIP: AverageLearner2D and AverageLearner1D

basnijholt:
## ([original merge request on GitLab](https://gitlab.kwant-project.org/qt/adaptive/merge_requests/70))

_opened by Bas Nijholt ([@basnijholt](https://gitlab.kwant-project.org/basnijholt)) at 2018-06-05T21:40:00.078Z_

This merge request implements a Learner2D that can learn averages on the points, the `AverageLearner2D`.

When chosing points the learner can either
* add more values at an existing points
* add more triangles

The learner compares the loss of potential new triangles with the standard error of an existing point.

The relative importance of both can be adjusted by a hyperparameter `learner.weight`.
From the doc-string:
>When `weight > 1` adding more points to existing points will be
            prioritized (making the standard error of a point more imporant,)
            otherwise adding new triangles will be prioritized (making the 
            loss of a triangle more important.)

All tests that pass for the `Learner2D` currently pass for the `AvererageLearner2D` too.

Run with:

```python
import adaptive
adaptive.notebook_extension()

def ring(xy_seed):
    import numpy as np
    import random
    (x, y), seed = xy_seed
    random.seed(seed)
    noise =  0.5 * (random.random() - 1)
    a = 0.2
    return x + np.exp(-(x**2 + y**2 - 0.75**2)**2/a**4) + noise

learner = adaptive.AverageLearner2D(ring, bounds=[(-1, 1), (-1, 1)], weight=.1)
runner = adaptive.Runner(learner, goal=lambda l: l.loss() < 0.01, log=True)
runner.live_info()
```

which results in:

```
>>> print(learner.mean_values_per_point()) 
5.2737642585551328
```
and

Screenshot_2018-09-27_at_18.03.37

### DIFF
--- a/adaptive/learner/average_mixin.py
+++ b/adaptive/learner/average_mixin.py
@@ -1,0 +1,155 @@
+# -*- coding: utf-8 -*-
+
+from math import sqrt
+import sys
+
+import numpy as np
+
+from .learner1D import Learner1D
+
+
+inf = sys.float_info.max
+
+
+class AverageMixin:
+    @property
+    def data(self):
+        return {k: sum(v.values()) / len(v) for k, v in self._data.items()}
+
+    @property
+    def data_sem(self):
+        return {k: self.standard_error(v.values())
+            for k, v in self._data.items()}
+
+    def standard_error(self, lst):
+        n = len(lst)
+        if n < self.min_values_per_point:
+            return inf
+        sum_f_sq = sum(x**2 for x in lst)
+        mean = sum(x for x in lst) / n
+        numerator = sum_f_sq - n * mean**2
+        if numerator < 0:
+            # This means that the numerator is ~ -1e-15
+            return 0
+        std = sqrt(numerator / (n - 1))
+        return std / sqrt(n)
+
+    def mean_values_per_point(self):
+        return np.mean([len(x.values()) for x in self._data.values()])
+
+    def get_seed(self, point):
+        _data = self._data.get(point, {})
+        pending_seeds = self.pending_points.get(point, set())
+        seed = len(_data) + len(pending_seeds)
+        if seed in _data or seed in pending_seeds:
+            # means that the seed already exists, for example
+            # when '_data[point].keys() | pending_points[point] == {0, 2}'.
+            return (set(range(seed)) - pending_seeds - _data.keys()).pop()
+        return seed
+
+    def loss_per_existing_point(self):
+        scale = self.value_scale()
+
+        points = []
+        loss_improvements = []
+        for p, sem in self.data_sem.items():
+            points.append((p, self.get_seed(p)))
+            N = self.n_values(p)
+            sem_improvement = (1 - sqrt(N - 1) / sqrt(N)) * sem
+            loss_improvement = self.weight * sem_improvement / scale
+            loss_improvements.append(loss_improvement)
+        return points, loss_improvements
+
+    def _add_to_pending(self, point):
+        x, seed = self.unpack_point(point)
+        if x not in self.pending_points:
+            self.pending_points[x] = set()
+        self.pending_points[x].add(seed)
+
+    def _remove_from_to_pending(self, point):
+        x, seed = self.unpack_point(point)
+        if x in self.pending_points:
+            self.pending_points[x].discard(seed)
+            if not self.pending_points[x]:
+                # pending_points[x] is now empty so delete the set()
+                del self.pending_points[x]
+
+    def _add_to_data(self, point, value):
+        x, seed = self.unpack_point(point)
+        if x not in self._data:
+            self._data[x] = {}
+        self._data[x][seed] = value
+
+    def ask(self, n, tell_pending=True):
+        """Return n points that are expected to maximally reduce the loss."""
+        points, loss_improvements = self._ask_points_without_adding(n)
+        loss_improvements = self._normalize_new_points_loss_improvements(
+            points, loss_improvements)
+
+        p, l = self.loss_per_existing_point()
+        l = self._normalize_existing_points_loss_improvements(p, l)
+        points += p
+        loss_improvements += l
+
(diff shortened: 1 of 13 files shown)


basnijholt:
_originally posted by Joseph Weston ([@jbweston](https://gitlab.kwant-project.org/jbweston)) at 2018-06-07T08:11:26.307Z on [GitLab](https://gitlab.kwant-project.org/qt/adaptive/merge_requests/70)_

Would the `AverageLearner2D` benefit from having `Learner2D` as its base class? Seems there's a lot of replicated code

basnijholt:
_originally posted by Bas Nijholt ([@basnijholt](https://gitlab.kwant-project.org/basnijholt)) at 2018-06-08T01:13:52.123Z on [GitLab](https://gitlab.kwant-project.org/qt/adaptive/merge_requests/70)_

That was my plan too, but I'll while working on it, it's easier to leave the code in one file. When it's done I'll make it a subclass of `Learner2D`.

basnijholt:
_originally posted by Bas Nijholt ([@basnijholt](https://gitlab.kwant-project.org/basnijholt)) at 2018-09-26T15:22:56.303Z on [GitLab](https://gitlab.kwant-project.org/qt/adaptive/merge_requests/70)_

This is needed to be able to do `super().__init__(function, bounds, loss_per_triangle)` in the `AverageLearner2D`.

basnijholt:
_originally posted by Bas Nijholt ([@basnijholt](https://gitlab.kwant-project.org/basnijholt)) at 2018-09-26T15:26:05.616Z on [GitLab](https://gitlab.kwant-project.org/qt/adaptive/merge_requests/70)_

The logic in the line below needs to be editted.

Right now we are literally comparing a `loss_improvement` with a `standard_error`.

We should normalize them in some way.

adaptive/learner/average_mixin.py
```diff
@@ -0,0 +1,202 @@
...
+        mean_values_per_neighbor = self._mean_values_per_neighbor(neighbors)
+
+        def needs_more_data(p):
+            return mean_values_per_neighbor[p] > 1.5 * self.n_values(p)
+
+        return [inf if needs_more_data(p) else loss
+            for (p, seed), loss in zip(points, loss_improvements)]
+
+
+def add_average_mixin(cls):
```
akhmerov: I'm not sure whether this is clear enough; feels like way too much magic. And from what I see, the usual inheritance should be good enough.

basnijholt: > the usual inheritance should be good enough

That was not the case. I don't remember the exact details but there were issues with the method resolution order, there was no other option than doing it like this.

If you want I could find out why exactly.

adaptive/learner/average_mixin.py
```diff
@@ -0,0 +1,202 @@
...
+             '_normalize_existing_points_loss_improvements',
+             '_mean_values_per_neighbor')
+
+    for name in names:
+        setattr(cls, name, getattr(AverageMixin, name))
+
+    return cls
+
+
+class DataPoint(dict):
```
akhmerov: Isn't this an `AverageLearner`?

basnijholt: Not really, it's similar.

It is really only a container that keeps track of the sum and sum of the squared values, such that it doesn't need to do the sum each time.

basnijholt: We can use it there too actually. I've added it in the `AverageLearner`.

basnijholt:
@anton I have implemented what you suggested in chat:
> It's not just that: the two options that make sense are:
> * increase the number of samples in a point by a fixed fraction (e.g. 1.1)
> * add a new point with the number of samples that's comparable to the number of samples in the neighboring points.

The problem now is that once a point has a lot of "seeds", increasing the number of seeds by 10% will give a big loss improvement, probably the biggest, so the number of values at that point will grow very big. Conceptually this shouldn't happen, so I probably made a mistake in the following method:
```python
    def loss_per_existing_point(self):
        """Increase the number of seeds by 10%."""
        if len(self.data) < 4:
            return [], []
        scale = self.value_scale()
        points = []
        loss_improvements = []

        neighbors = self._get_neighbor_mapping_existing_points()
        mean_values_per_neighbor = self._mean_values_per_neighbor(neighbors)

        for p, sem in self.data_sem.items():
            n_neighbors = mean_values_per_neighbor[p]
            N = self.n_values(p)
            n_more = int(1.1 * N)  # increase the amount of points by 10%
            points.append((p, n_more))
            # This is the improvement considering we will add
            # n_more seeds to the stack.
            sem_improvement = (1 / sqrt(N) - 1 / sqrt(N + n_more)) * sem
            loss_improvement = self.weight * sem_improvement / scale  # XXX: Do I need to divide by the scale?
            loss_improvements.append(loss_improvement)
        return points, loss_improvements
```

akhmerov:
> The problem now is that once a point has a lot of "seeds", increasing the number of seeds by 10% will give a big loss improvement, probably the biggest, so the number of values at that point will grow very big.

If you increase the number of points by 10%, the rms at the point drops by 5%; why would this be the biggest loss improvement?

basnijholt:
@akhmerov it is fixed now, there were several bugs.

Everything seems to work now.

I've also renamed the `weight` to `average_priority`.

basnijholt:
I've noticed that the `AverageLearner1D/2D` aren't working for `BalancingLearners` because of the `ask(..., tell_pending=False)` when using `bal_learner.strategy = 'loss_improvements'`.

basnijholt:
I've added a cool plotting feature, on hovering over the points it displays extra information:
image

basnijholt:
The failing test `test_saving[AverageLearner1D-random_linear_with_peak-learner_kwargs6]` is because the y_scale not properly being updated.

PR #151, opened 2019-02-14
handle NaN losses and add a test, closes #145

### LINKED ISSUE #145: Returning np.nan breaks the 1D learner
From commit: 4381a9e52faef7332b2ad0d3354ebb44616df7d9 onward,
returning (a np.array of) np.nan cancels the runner.
Example code, to compare the working commit, f38d20fe2899582c46fcc23f0612242cce6904bd,
with the not working one, 4381a9e52faef7332b2ad0d3354ebb44616df7d9.
```python
import adaptive
adaptive.notebook_extension()
import numpy as np
def f(x):
    if 0 < x < 1:
        return x
    else:
        return np.nan

learner = adaptive.Learner1D(f, 0.1*np.array([-1, 1]))
runner = adaptive.Runner(learner)
runner.live_info()
```

basnijholt:
This reverts the behavior that we've decided upon a long time ago.

The bug was introduced a few months ago.

### DIFF
--- a/adaptive/learner/learner1D.py
+++ b/adaptive/learner/learner1D.py
@@ -684,7 +684,7 @@ def finite_loss(ival, loss, x_scale):
     sort intervals that have infinite loss."""
     # If the loss is infinite we return the
     # distance between the two points.
-    if math.isinf(loss):
+    if math.isinf(loss) or math.isnan(loss):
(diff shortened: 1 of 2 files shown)


jbweston:
LGTM

PR #160, opened 2019-03-16
Distribute first points in a BalancingLearner

### LINKED ISSUE #159: BalancingLearner puts all points in the first child-learner when asking for points with no data present
```python
def h(x, offset=0):
    a = 0.01
    return x + a**2 / (a**2 + (x - offset)**2)

learners = [adaptive.Learner1D(partial(h, offset=random.uniform(-1, 1)),
            bounds=(-1, 1)) for i in range(10)]

bal_learner = adaptive.BalancingLearner(learners)

bal_learner.ask(10)
```
```
([(0, -1),
  (0, 1),
  (0, 0.0),
  (0, -0.5),
  (0, 0.5),
  (0, -0.75),
  (0, -0.25),
  (0, 0.25),
  (0, 0.75),
  (0, -0.875)],
 [inf, inf, inf, inf, inf, inf, inf, inf, inf, inf])
```

Instead, it should balance the points over all learners. This happens because the loss improvements are `inf` for all points.

A temporary solution would be to force all learners to have the same amount of points by using:
```
bal_learner = adaptive.BalancingLearner(learners, strategy='npoints')
```

basnijholt:
This fixes #159.

@akhmerov could you take a look at this?

### DIFF
--- a/adaptive/learner/balancing_learner.py
+++ b/adaptive/learner/balancing_learner.py
@@ -77,7 +77,7 @@ class BalancingLearner(BaseLearner):
         # pickle the whole learner.
         self.function = partial(dispatch, [l.function for l in self.learners])
 
-        self._points = {}
+        self._ask_cache = {}
         self._loss = {}
         self._pending_loss = {}
         self._cdims_default = cdims
@@ -122,10 +122,10 @@ class BalancingLearner(BaseLearner):
             points_per_learner = []
             for index, learner in enumerate(self.learners):
                 # Take the points from the cache
-                if index not in self._points:
-                    self._points[index] = learner.ask(
+                if index not in self._ask_cache:
+                    self._ask_cache[index] = learner.ask(
                         n=1, tell_pending=False)
-                points, loss_improvements = self._points[index]
+                points, loss_improvements = self._ask_cache[index]
 
                 priority = (loss_improvements[0], -npoints[index])
                 improvements_per_learner.append(priority)
@@ -154,9 +154,9 @@ class BalancingLearner(BaseLearner):
             npoints[index] += 1
 
             # Take the points from the cache
-            if index not in self._points:
-                self._points[index] = self.learners[index].ask(n=1)
-            points, loss_improvements = self._points[index]
+            if index not in self._ask_cache:
+                self._ask_cache[index] = self.learners[index].ask(n=1)
+            points, loss_improvements = self._ask_cache[index]
 
             chosen_points.append((index, points[0]))
             chosen_loss_improvements.append(loss_improvements[0])
@@ -171,9 +171,9 @@ class BalancingLearner(BaseLearner):
         while n_left > 0:
             index = np.argmin(npoints)
             # Take the points from the cache
-            if index not in self._points:
-                self._points[index] = self.learners[index].ask(n=1)
-            points, loss_improvements = self._points[index]
+            if index not in self._ask_cache:
+                self._ask_cache[index] = self.learners[index].ask(n=1)
+            points, loss_improvements = self._ask_cache[index]
             npoints[index] += 1
             n_left -= 1
             chosen_points.append((index, points[0]))
@@ -190,14 +190,14 @@ class BalancingLearner(BaseLearner):
 
     def tell(self, x, y):
         index, x = x
-        self._points.pop(index, None)
+        self._ask_cache.pop(index, None)
         self._loss.pop(index, None)
         self._pending_loss.pop(index, None)
         self.learners[index].tell(x, y)
 
     def tell_pending(self, x):
         index, x = x
-        self._points.pop(index, None)
+        self._ask_cache.pop(index, None)
         self._loss.pop(index, None)
         self.learners[index].tell_pending(x)
 
--- a/adaptive/learner/balancing_learner.py
+++ b/adaptive/learner/balancing_learner.py
@@ -121,6 +121,7 @@ class BalancingLearner(BaseLearner):
             improvements_per_learner = []
             points_per_learner = []
             for index, learner in enumerate(self.learners):
+                # Take the points from the cache
                 if index not in self._points:
                     self._points[index] = learner.ask(
                         n=1, tell_pending=False)
@@ -155,27 +156,35 @@ class BalancingLearner(BaseLearner):
                          + len(l.pending_points))
                        for i, l in enumerate(self.learners)]
             priority = zip(losses, npoints)
-            index, (_, _) = max(enumerate(priority), key=itemgetter(1))
+            index = max(enumerate(priority), key=itemgetter(1))[0]
             npoints_per_learner[index] += 1
-            points, loss_improvements = self.learners[index].ask(1)
+
+            # Take the points from the cache
+            if index not in self._points:
+                self._points[index] = self.learners[index].ask(n=1)
+            points, loss_improvements = self._points[index]
+
             chosen_points.append((index, points[0]))
             chosen_loss_improvements.append(loss_improvements[0])
         return chosen_points, chosen_loss_improvements
 
     def _ask_and_tell_based_on_npoints(self, n):
-        points = []
-        loss_improvements = []
+        chosen_points = []
+        chosen_loss_improvements = []
         npoints = [l.npoints + len(l.pending_points)
                    for l in self.learners]
         n_left = n
         while n_left > 0:
-            i = np.argmin(npoints)
-            xs, ls = self.learners[i].ask(1)
-            npoints[i] += 1
+            index = np.argmin(npoints)
+            # Take the points from the cache
+            if index not in self._points:
+                self._points[index] = self.learners[index].ask(n=1)
+            points, loss_improvements = self._points[index]
+            npoints[index] += 1
             n_left -= 1
-            points.append((i, xs[0]))
-            loss_improvements.append(ls[0])
-        return points, loss_improvements
+            chosen_points.append((index, points[0]))
+            chosen_loss_improvements.append(loss_improvements[0])
+        return chosen_points, chosen_loss_improvements
 
     def ask(self, n, tell_pending=True):
         """Chose points for learners."""
--- a/adaptive/learner/balancing_learner.py
+++ b/adaptive/learner/balancing_learner.py
@@ -113,40 +113,34 @@ class BalancingLearner(BaseLearner):
                 ' strategy="npoints" is implemented.')
 
     def _ask_and_tell_based_on_loss_improvements(self, n):
-        chosen_points = []
-        chosen_loss_improvements = []
-        npoints = [l.npoints + len(l.pending_points)
-                   for l in self.learners]
+        selected = []  # tuples ((learner_index, point), loss_improvement)
+        npoints = [l.npoints + len(l.pending_points) for l in self.learners]
         for _ in range(n):
-            improvements_per_learner = []
-            points_per_learner = []
+            to_select = []
             for index, learner in enumerate(self.learners):
                 # Take the points from the cache
                 if index not in self._ask_cache:
                     self._ask_cache[index] = learner.ask(
                         n=1, tell_pending=False)
                 points, loss_improvements = self._ask_cache[index]
-
-                priority = (loss_improvements[0], -npoints[index])
-                improvements_per_learner.append(priority)
-                points_per_learner.append((index, points[0]))
+                to_select.append(
+                    ((index, points[0]),
+                     (loss_improvements[0], -npoints[index]))
+                )
 
             # Choose the optimal improvement.
             (index, point), (loss_improvement, _) = max(
-                zip(points_per_learner, improvements_per_learner),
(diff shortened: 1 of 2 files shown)


akhmerov:
This fix seems to be only partial: consider for example the case when one learner is much cheaper than all the others. Then as soon as one point arrives, the balancing learner will request all points from another random learner in the queue.

If I understand correctly, an empty learner has both an infinite loss and an infinite loss improvement, however some non-empty learners also have the same loss and loss improvement. In that case the problem would arise as well.

Therefore it seems to suggest that a better fix would be to sort on `(-loss, n_points)` (or loss improvement, depending on the balancing learner setting).

basnijholt:
@akhmerov, good suggestion! 

I've done this in 6e172ef85228e5f4469488fe07d6ae0773b4b498. This is what you meant right?

I'll also do it in `_ask_and_tell_based_on_loss`.

adaptive/learner/balancing_learner.py
```diff
@@ -112,25 +112,33 @@ def strategy(self, strategy):
...
-            x, l = max(zip(pairs, improvements_per_learner),
-                       key=itemgetter(1))
-            points.append(x)
-            loss_improvements.append(l)
-            self.tell_pending(x)
-
-        return points, loss_improvements
+                points, loss_improvements = self._points[index]
+                npoints = npoints_per_learner[index] + learner.npoints
+                sort_tuple = (loss_improvements[0], -npoints)
```
akhmerov: `priority`?

adaptive/learner/balancing_learner.py
```diff
@@ -112,25 +112,33 @@ def strategy(self, strategy):
...
+                improvements_per_learner.append(sort_tuple)
+                points_per_learner.append((index, points[0]))
+
+            # Chose the optimal improvement.
+            (index, point), (loss_improvement, _) = max(
+                zip(points_per_learner, improvements_per_learner),
+                key=itemgetter(1))
+            npoints_per_learner[index] += 1
```
akhmerov: Does this ever get decremented?

basnijholt: No it won't.

basnijholt: No it won't.

akhmerov: But then what does it mean? Above I see `npoints_per_learner[index] + learner.npoints`

basnijholt: learner.npoints doesn't change when we add a point to chosen_points, so without this all suggested points would come from the same learner.

basnijholt: learner.npoints doesn't change when we add a point to chosen_points, so without this all suggested points would come from the same learner.

akhmerov: But shouldn't we decrement this when a point is added to the learner? Also what is the meaning of this variable? It's not at all clear from how it's used, and seems to duplicate `learner.npoints`.

basnijholt: but `learner.npoints` doesn't increase when adding something to `tell_pending`.

basnijholt: So in this loop `for index, learner in enumerate(self.learners):` we ask a point from each learner. After this we can pick the `learner` and `point` with the highest `loss_improvement`, this means that in our `chosen_points` list we will have one point for the `learner` we chose. However, `learner.npoints` wouldn't have increased at this point in time! So the next time we execute this loop we need to take the `chosen_points` into account.

akhmerov: Ah! I overlooked that this isn't an attribute.

akhmerov:
My problem right now is that the logic is rather nontrivial, and it's hard to follow especially due to very similar variable names. Comments might also help.

adaptive/learner/balancing_learner.py
```diff
@@ -111,38 +111,57 @@ def strategy(self, strategy):
             raise ValueError(
                 'Only strategy="loss_improvements", strategy="loss", or'
                 ' strategy="npoints" is implemented.')
+        self._points = {}  # reset the cache
 
     def _ask_and_tell_based_on_loss_improvements(self, n):
-        points = []
-        loss_improvements = []
+        chosen_points = []
```
akhmerov:
```python
selected = []  # tuples (learner_index, point) 
```

adaptive/learner/balancing_learner.py
```diff
@@ -111,38 +111,57 @@ def strategy(self, strategy):
...
+        chosen_loss_improvements = []
+        npoints_per_learner = defaultdict(int)
```
akhmerov: Rename to `n_selected` or something like that

Add a comment along the lines of "`same as Counter(i[0] for i in selected)`"

adaptive/learner/balancing_learner.py
```diff
@@ -111,38 +111,57 @@ def strategy(self, strategy):
             raise ValueError(
                 'Only strategy="loss_improvements", strategy="loss", or'
                 ' strategy="npoints" is implemented.')
+        self._points = {}  # reset the cache
 
     def _ask_and_tell_based_on_loss_improvements(self, n):
-        points = []
-        loss_improvements = []
+        chosen_points = []
+        chosen_loss_improvements = []
```
akhmerov: Just `loss_improvements`?

adaptive/learner/balancing_learner.py
```diff
@@ -111,38 +111,57 @@ def strategy(self, strategy):
...
-                pairs.append((index, point[0]))
-            x, l = max(zip(pairs, improvements_per_learner),
-                       key=itemgetter(1))
-            points.append(x)
-            loss_improvements.append(l)
-            self.tell_pending(x)
-
-        return points, loss_improvements
+                points, loss_improvements = self._points[index]
+                npoints = (npoints_per_learner[index]
```
akhmerov: `total_points` or just `total`

adaptive/learner/balancing_learner.py
```diff
@@ -111,38 +111,57 @@ def strategy(self, strategy):
...
-                improvements_per_learner.append(loss_improvement[0])
-                pairs.append((index, point[0]))
-            x, l = max(zip(pairs, improvements_per_learner),
-                       key=itemgetter(1))
-            points.append(x)
-            loss_improvements.append(l)
-            self.tell_pending(x)
-
-        return points, loss_improvements
+                points, loss_improvements = self._points[index]
```
akhmerov: This is really confusing and seems like we've chosen a bad name for `self._points`: from that name it's not at all clear that loss improvements are stored, neither it is clear what these `self._points` actually are. All points ever requested?

basnijholt:
I found another issue that I fixed. 

I've simplified it now and I think the code is understandable, do you agree?

adaptive/learner/balancing_learner.py
akhmerov: Overall we should make a clear distinction between `point` and `(learner_index, point)`. Conceptually these are the same (the latter is a point from a balancing learner perspective). This makes them easy to confuse, and therefore a clear name separation is very useful.

adaptive/learner/balancing_learner.py
basnijholt: It's an cache for `ask`, so maybe `_ask_cache`?

akhmerov: Yes, that's much better.

adaptive/learner/balancing_learner.py
basnijholt: The code is no longer the same, so I resolved it.

adaptive/learner/balancing_learner.py
basnijholt: I call `loss_improvements` what is returned be `ask`, so always `loss_improvements = ask(n=1)[1]`.

adaptive/learner/balancing_learner.py
basnijholt: Maybe we could make `selected` like:

`selected = []  # tuples ((learner_index, point), loss_improvement) `

akhmerov: Indeed, however shouldn't it be flattened as well then?

basnijholt: Yes, like I did in the last commit?

adaptive/learner/balancing_learner.py
```diff
@@ -113,50 +113,67 @@ def strategy(self, strategy):
                 ' strategy="npoints" is implemented.')
 
     def _ask_and_tell_based_on_loss_improvements(self, n):
-        points = []
-        loss_improvements = []
+        selected = [] # tuples ((learner_index, point), loss_improvement)
+        npoints = [l.npoints + len(l.pending_points) for l in self.learners]
```
akhmerov: Maybe `total_points` (to indicate that it's evaluated and pending together)?

adaptive/learner/balancing_learner.py
```diff
@@ -113,50 +113,67 @@ def strategy(self, strategy):
...
         for _ in range(n):
             improvements_per_learner = []
-            pairs = []
+            points_per_learner = []
```
akhmerov: These aren't per learner (I'd only call something "per learner" if it's a list of length equal to the number of learners).

akhmerov: Are these `requested`?

adaptive/learner/balancing_learner.py
akhmerov: Yes, but what about flattening this data structure as well?

akhmerov: NB: PEP8 suggests a double space before a trailing comment.

adaptive/learner/balancing_learner.py
basnijholt: I've changed it. What do you think?

adaptive/learner/balancing_learner.py
basnijholt: Done.

adaptive/learner/balancing_learner.py
basnijholt: Fixed!

adaptive/learner/balancing_learner.py
basnijholt: Done

adaptive/learner/balancing_learner.py
akhmerov: Why not also flatten, so `(learner_index, point, loss_improvement)`?

akhmerov: Disregard that; it's the format the runner expects.

adaptive/learner/balancing_learner.py
```diff
@@ -113,54 +113,72 @@ def strategy(self, strategy):
...
-            self.tell_pending(x)
-
+                points, loss_improvements = self._ask_cache[index]
+                to_select.append(
+                    ((index, points[0]),
+                     (loss_improvements[0], -total_points[index]))
+                )
+
+            # Choose the optimal improvement.
+            (index, point), (loss_improvement, _) = max(
```
akhmerov: Shouldn't we remove the point from `_ask_cache` around here?

basnijholt: Very well spotted!

I've added `self.tell_pending((index, points[0]))`.

adaptive/learner/balancing_learner.py
akhmerov: Much better now. I think a reader can now understand what's happening without looking elsewhere.

adaptive/tests/test_balancing_learner.py
```diff
@@ -21,3 +22,34 @@ def test_balancing_learner_loss_cache():
     bl = BalancingLearner([learner])
     assert bl.loss(real=False) == pending_loss
     assert bl.loss(real=True) == real_loss
+
+
+def test_distribute_first_points_over_learners():
+    for strategy in ['loss', 'loss_improvements', 'npoints']:
```
akhmerov: `pytest.mark.parameterize`? (same below)

adaptive/tests/test_balancing_learner.py
```diff
@@ -21,3 +22,34 @@ def test_balancing_learner_loss_cache():
...
+        learners = [Learner1D(lambda x: x, bounds=(-1, 1)) for i in range(10)]
+        learner = BalancingLearner(learners, strategy=strategy)
+        points, _ = learner.ask(100)
```
akhmerov: We should also test for a failure mode when e.g. 1 or 3 points are fed into a learner.

adaptive/tests/test_balancing_learner.py
basnijholt: Good one. Done!

adaptive/tests/test_balancing_learner.py
basnijholt: Done in 6e3e9852e828ee5f5d016f26a4d1e2453c606aaf.

PR #161, opened 2019-03-17
import Iterable and Sized from collections.abc

basnijholt:
Because of the following DeprecationWarning:
```
DeprecationWarning: Using or importing the ABCs from 'collections'
instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```


akhmerov:
LGTM

PR #168, opened 2019-03-19
Temporarily fix docs

basnijholt:
This is not a proper fix but before I discovered how to fix it correctly this might do.

After this plot.ly and HoloMaps will correctly show.


akhmerov:
Please add [the license](https://github.com/rtfd/sphinx_rtd_theme/blob/master/LICENSE) to the file.

akhmerov:
Is it possible to verify this works without building locally? Is there a preview on rtd?

basnijholt:
I just build it locally but I removed most files to speed it up. The LearnerND tutorial is the only file you need to check.

PR #170, opened 2019-03-21
Add logo to the documentation

basnijholt:
This adds a logo to the documentation.

`python logo.py` generates the image.

Alternatively, we could also generate the logo in CI.

image


jbweston:
LGTM! Maybe consider adding a comment about how to run the script (i.e. make sure to run as `./logo.py` from the documentation directory)

jbweston:
We should also automatically re-run when `logo.py` changes. I didn't manage to fine the right incantation for the makefile yet.

basnijholt:
Great that you like it Joe!

@akhmerov actually thought that we should pick something with just a random bunch of points such that the triangulation can be seen better. However, I don't agree with that, Adaptive is not about triangulation but about "learning" a function, so I think the logo should represent that. The circle is very recognizable and I think it represents the concept of Adaptive really well.

jbweston:
> Adaptive is not about triangulation but about "learning" a function, so I think the logo should represent that.

The circle logo has grown on me also.

I'll wait for the checks to complete and then merge

jbweston:
The failure has nothing to do with the change so I'm tempted to just merge anyway.

PR #180, opened 2019-04-03
use setup(..., python_requires='>=3.6'), closes #179

### LINKED ISSUE #179: set python_requires in setup.py
See https://python3statement.org/practicalities/

python_requires will tell to pip ahead of time not to even _try_ to install on incompatible versions. That will avoid error message on triggered by the test in here https://github.com/python-adaptive/adaptive/blob/master/setup.py#L8-L10. 

it's too late for already released versions, but at least for the newest ones you could set that up.

basnijholt:
As by the suggestion of @Carreau in #179.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,7 @@ setup(
     name='adaptive',
     description='Adaptive parallel sampling of mathematical functions',
     version=version,
+    python_requires='>=3.6',
     url='https://adaptive.readthedocs.io/',
     author='Adaptive authors',
     license='BSD',


akhmerov:
I see the following in py27:

```python
pip install git+https://github.com/python-adaptive/adaptive.git@python_requires                                                                                                      (py27) 
Collecting git+https://github.com/python-adaptive/adaptive.git@python_requires
  Cloning https://github.com/python-adaptive/adaptive.git (to revision python_requires) to /tmp/pip-req-build-uRyzYI
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-uRyzYI/setup.py", line 19, in <module>
        version, cmdclass = get_version_and_cmdclass('adaptive')
      File "/tmp/pip-req-build-uRyzYI/setup.py", line 11, in get_version_and_cmdclass
        from importlib.util import module_from_spec, spec_from_file_location
    ImportError: No module named util
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-req-build-uRyzYI/
```

setup.py
```diff
@@ -5,11 +5,6 @@
 import sys
 
 
-if sys.version_info < (3, 6):
-    print('adaptive requires Python 3.6 or above.')
-    sys.exit(1)
-
-
```
Carreau: Both are not incompatible. This check is still useful if you have users with `pip < 9.0.1` (which is still often the case for Python 2 users).

basnijholt:
Thanks for checking, I've reverted that change.

akhmerov:
LGTM. pip still fails on 2.7 if installing from source, but that's an unrelated issue which we should fix separately.

PR #182, opened 2019-04-10
2D: do not return points outside the bounds, closes #181

### LINKED ISSUE #181: Learner2D.plot() returns NaN
Example:

```python
import adaptive
adaptive.notebook_extension()
import numpy as np

bounds = [(1, 1.7), (250, 550)]
learner = adaptive.Learner2D(lambda x: x[0] * x[1], bounds=bounds)
adaptive.runner.simple(learner, goal=lambda l: l.npoints >= 2000)

plot_data = learner.plot().Image.I.data
print(np.sum(np.isnan(plot_data)))
>>> 47
```
The issue is the precision in `Learner2D._scale` (which is called by `Learner2D.ip`, which itself is used to make the plot):

```python
points, values = learner._data_in_bounds()
points = learner._scale(points)
print(np.min(points[:, 0]), np.max(points[:, 0]))
>>> -0.5000000000000001 0.49999999999999983
```

Points lying exactly at `x = 0.5` will be outside of the convex hull of `points` and `scipy.interpolate.LinearNDInterpolator` [fills them with `nan`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.LinearNDInterpolator.html).

basnijholt:
@andrenmelo, this is the (a) solution.

@python-adaptive/core what do you think?

### DIFF
--- a/adaptive/learner/learner2D.py
+++ b/adaptive/learner/learner2D.py
@@ -543,7 +543,14 @@ class Learner2D(BaseLearner):
                 n = int(0.658 / sqrt(areas(ip).min()))
                 n = max(n, 10)
 
-            x = y = np.linspace(-0.5, 0.5, n)
+            # The bounds of the linspace should be (-0.5, 0.5) but because of
+            # numerical precision problems it could (for example) be
+            # (-0.5000000000000001, 0.49999999999999983), then any point at exact
+            # boundary would be outside of the domain. See #181.
+            x_min, y_min = np.min(ip.points, axis=0)
+            x_max, y_max = np.max(ip.points, axis=0)
+            x = np.linspace(x_min, x_max, n)
+            y = np.linspace(y_min, y_max, n)
             z = ip(x[:, None], y[None, :] * self.aspect_ratio).squeeze()
 
             if self.vdim > 1:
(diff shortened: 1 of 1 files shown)


akhmerov:
Here the numerical errors don't accumulate over time, and therefore we may simply choose a relative tolerance that is merely much higher than the double precision (e.g. `1e-13`), and not go over the complete interpolation each time.

Also as implemented, there is no guarantee that all the points that are selected are within the triangulation if e.g. the midpoint on the edge of the rectangle is outermost and the top or bottom are ε away from it.

basnijholt:
> Also as implemented, there is no guarantee that all the points that are selected are within the triangulation if e.g. the midpoint on the edge of the rectangle is outermost and the top or bottom are ε away from it.

I think that the rounding of the same number should always happen the same, so I don't think this could occur.

> Here the numerical errors don't accumulate over time, and therefore we may simply choose a relative tolerance that is merely much higher than the double precision (e.g. 1e-13), and not go over the complete interpolation each time.

So you just mean doing?
```python
eps = 1e-13
x = y = np.linspace(-0.5 + eps, 0.5 - eps, n)
```

akhmerov:
LGTM

PR #185, opened 2019-04-18
Add support for neighbours in loss computation in LearnerND

jhoofwijk:
This PR will make the LearnerND get the option to add more points in highly curved areas (similar to the Learner1D)

It also moves the `uses_nth_neighbor` function from the `learner1D` to the `base_learner` to avoid circular imports


basnijholt:
This looks great @JornHoofwijk, I'll soon go over your new changes! 😄

PR #188, opened 2019-04-25
BalancingLearner: add a "cycle" strategy, sampling the learners one by one

basnijholt:
This is useful for example for the `AverageLearner1D` and `2D` where sampling on `npoints` won't sample all learners equally, because that's not how `npoints` is defined there.

### DIFF
--- a/adaptive/learner/balancing_learner.py
+++ b/adaptive/learner/balancing_learner.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 from collections.abc import Iterable
 from contextlib import suppress
 from functools import partial
+import itertools
 from operator import itemgetter
 
 import numpy as np
@@ -53,7 +54,8 @@ class BalancingLearner(BaseLearner):
     strategy : 'loss_improvements' (default), 'loss', or 'npoints'
         The points that the `BalancingLearner` choses can be either based on:
         the best 'loss_improvements', the smallest total 'loss' of the
-        child learners, or the number of points per learner, using 'npoints'.
+        child learners, the number of points per learner, using 'npoints',
+        or by cycling through the learners one by one using 'cycle'.
         One can dynamically change the strategy while the simulation is
         running by changing the ``learner.strategy`` attribute.
 
@@ -90,10 +92,11 @@ class BalancingLearner(BaseLearner):
 
     @property
     def strategy(self):
-        """Can be either 'loss_improvements' (default), 'loss', or 'npoints'
-        The points that the `BalancingLearner` choses can be either based on:
-        the best 'loss_improvements', the smallest total 'loss' of the
-        child learners, or the number of points per learner, using 'npoints'.
+        """Can be either 'loss_improvements' (default), 'loss', 'npoints', or
+        'cycle'. The points that the `BalancingLearner` choses can be either
+        based on: the best 'loss_improvements', the smallest total 'loss' of
+        the child learners, the number of points per learner, using 'npoints',
+        or by going through all learners one by one using 'cycle'.
         One can dynamically change the strategy while the simulation is
         running by changing the ``learner.strategy`` attribute."""
         return self._strategy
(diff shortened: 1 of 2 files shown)


adaptive/learner/balancing_learner.py
```diff
@@ -176,6 +181,20 @@ def _ask_and_tell_based_on_npoints(self, n):
         points, loss_improvements = map(list, zip(*selected))
         return points, loss_improvements
 
+    def _ask_and_tell_based_on_cycle(self, n):
+        if not hasattr(self, '_cycle'):
```
basnijholt: This could also just be set in the `__init__`.

akhmerov:
I think this is rather an indication of `npoints` being inconsistently defined in `AverageLearner`. We already have a strategy that aims to make the amount of data per learner equal.

If this is a preferable implementation, we should rather remove the other strategy. However it'd make most sense if `npoints` is the amount of data points.

basnijholt:
But `cycle` does something different than `npoints`, you might, for example, run a `BalancingLearner` for `loss_improvements` and then just increase the number of points per learner with `cycle`, instead of making the number of points equal for all learners using `npoints`.

akhmerov:
I see, that's different indeed. Is that a behavior we would want to support?

It is also different from essentially everything else in adaptive: it tells the learner what to do and now what goal to reach.

adaptive/learner/balancing_learner.py
```diff
@@ -53,7 +54,8 @@ class BalancingLearner(BaseLearner):
     strategy : 'loss_improvements' (default), 'loss', or 'npoints'
```
jbweston: add 'cycle' here

adaptive/learner/balancing_learner.py
```diff
@@ -107,10 +110,12 @@ def strategy(self, strategy):
             self._ask_and_tell = self._ask_and_tell_based_on_loss
         elif strategy == "npoints":
             self._ask_and_tell = self._ask_and_tell_based_on_npoints
+        elif strategy == "cycle":
+            self._ask_and_tell = self._ask_and_tell_based_on_cycle
```
jbweston: I would add `self._cycle` here, rather than in `_ask_and_tell`

adaptive/learner/balancing_learner.py
```diff
@@ -173,6 +178,20 @@ def _ask_and_tell_based_on_npoints(self, n):
         points, loss_improvements = map(list, zip(*selected))
         return points, loss_improvements
 
+    def _ask_and_tell_based_on_cycle(self, n):
+        if not hasattr(self, "_cycle"):
+            self._cycle = itertools.cycle(range(len(self.learners)))
```
jbweston: do this in `__init__` as I commented above.

basnijholt: But you wrote to do in the strategy property setter?

basnijholt:
@jbweston I think I've done what you suggested (up to https://github.com/python-adaptive/adaptive/pull/188#discussion_r301834980).

Merge if you like.

adaptive/learner/balancing_learner.py
```diff
@@ -107,10 +110,13 @@ def strategy(self, strategy):
             self._ask_and_tell = self._ask_and_tell_based_on_loss
         elif strategy == "npoints":
             self._ask_and_tell = self._ask_and_tell_based_on_npoints
+        elif strategy == "cycle":
+            self._ask_and_tell = self._ask_and_tell_based_on_cycle
+            self._cycle = itertools.cycle(range(len(self.learners)))
```
jbweston: Now the cycle will be reset every time the strategy is set dynamically. I'm not sure what the best thing to do here is.

basnijholt: That might be the intention? We could also just put in in `__init__`...

PR #193, opened 2019-05-08
Add a SequenceLearner

basnijholt:
Adaptive has really great features to trivially parallelize and auto-save your simulation, however, sometimes you just can't formulate your problem in one of the `Learner`s. Then just for this single problem, one needs to write a lot of code to basically do what Adaptive can already do for you.

Therefore I've implemented a `SequenceLearner` that doesn't do a whole lot of fancy things, except for being a `BaseLearner` child.

With this one can easily run, `save` (and periodically save), `load`, `live_info`, and `live_plot` one's progress (if one implements a simple `plotter` function).

It's great when running simulations that might crash because if one uses periodic saving, the learner just picks up where it was before it crashed (in contrary to `ipyparallel.Client().map` where one needs to implement all those features).

@jbweston and @akhmerov, what do you guys think? If this doesn't belong in Adaptive I'll put it in another package.

Example:
```python
def f(x):
    x = int(x)
    return x ** 2

seq = range(1000)
learner = SequenceLearner(f, seq)
adaptive.runner.simple(learner, SequenceLearner.done)
# that goal is same as `lambda l: l.done()`
```

```python
def plotter(learner):
    data = learner.data if learner.data else []
    return hv.Scatter(data)

learner = SequenceLearner(f, seq)
runner = adaptive.Runner(learner, SequenceLearner.done)
runner.live_plot(plotter=plotter)
```

```python
def g(xy):
    x, y = xy
    return x ** 2 + y**2

seq = np.random.rand(1000, 2)
learner = SequenceLearner(g, seq)
runner = adaptive.Runner(learner, SequenceLearner.done)
runner.live_info()
```

### DIFF
--- a/adaptive/learner/sequence_learner.py
+++ b/adaptive/learner/sequence_learner.py
@@ -1,0 +1,82 @@
+from copy import copy
+import sys
+
+from adaptive.learner.base_learner import BaseLearner
+
+inf = sys.float_info.max
+
+
+def ensure_hashable(x):
+    try:
+        hash(x)
+        return x
+    except TypeError:
+        return tuple(x)
+
+
+class SequenceLearner(BaseLearner):
+    def __init__(self, function, sequence):
+        self.function = function
+        self._to_do_seq = {ensure_hashable(x) for x in sequence}
+        self._npoints = len(sequence)
+        self.sequence = copy(sequence)
+        self.data = {}
+        self.pending_points = set()
+
+    def ask(self, n, tell_pending=True):
+        points = []
+        loss_improvements = []
+        i = 0
+        for point in self._to_do_seq:
+            if i > n:
+                break
+            points.append(point)
+            loss_improvements.append(inf / self._npoints)
+            i += 1
+
+        if tell_pending:
+            for p in points:
+                self.tell_pending(p)
+
+        return points, loss_improvements
+
+    def _get_data(self):
+        return self.data
+
+    def _set_data(self, data):
+        if data:
+            self.tell_many(*zip(*data.items()))
+
+    def loss(self, real=True):
+        if not (self._to_do_seq or self.pending_points):
+            return 0
+        else:
+            npoints = self.npoints + (0 if real else len(self.pending_points))
+            return inf / npoints
+
+    def remove_unfinished(self):
+        for p in self.pending_points:
+            self._to_do_seq.add(p)
+        self.pending_points = set()
+
+    def tell(self, point, value):
+        self.data[point] = value
+        self.pending_points.discard(point)
+        self._to_do_seq.discard(point)
+
+    def tell_pending(self, point):
+        self.pending_points.add(point)
+        self._to_do_seq.discard(point)
+
+    def done(self):
+        return not self._to_do_seq and not self.pending_points
+
+    def result(self):
+        """Get back the data in the same order as ``sequence``."""
+        if not self.done():
+            raise Exception("Learner is not yet complete.")
+        return [self.data[ensure_hashable(x)] for x in self.sequence]
+
+    @property
+    def npoints(self):
+        return len(self.data)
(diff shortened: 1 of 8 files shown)


basnijholt:
I did some testing to see how Adaptive compares with `ipyparallel`:
```
import adaptive
adaptive.notebook_extension()
import asyncio
import ipyparallel
c = ipyparallel.Client()
lview = c.load_balanced_view()
```

```
def f(x, wait=False):
    if wait:
        import time
        time.sleep(1)
    return 1

seq = list(range(5000))
%time res = lview.map_sync(f, seq)
```

Using a local `ipcluster` with 48 local engines:
- `wait=False` 2.77 s (`lview`) vs 2.82 s (`adaptive.Runner`) 
- `wait=True` 22.5 (`lview`) vs 22.8 s (`adaptive.Runner`) 

On the cluster, 100 engines on the `hpc05` (cluster) and the notebook on `io` (different machine outside of the cluster):
- `wait=False` 13.8 s (`lview`) vs 16.4 s (`adaptive.Runner`) 
- `wait=True` 55 s (`lview`) vs 51 s (`adaptive.Runner`) 

Surprisingly. somehow Adaptive is faster in the last case!

basnijholt:
Until this is merged the `SequenceLearner` also lives here https://github.com/basnijholt/adaptive-scheduler/blob/master/adaptive_scheduler/sequence_learner.py.

The requirement for the sequence to be hashable might be too strict. We could instead make `self._to_do_seq` into `self._to_do_indices` and pass the index to the function which would then internally be wrapped with `functools.partial` and call the function with the right parameters, similar to what we do in the `BalancingLearner`.

akhmerov:
While I don't think this is a useful addition to the library, it also does not extend the API and shouldn't break anything (if implemented correctly). Therefore it may be a useful example to those wishing to implement their own learner (or those, indeed, who find the non-adaptive part of adaptive relevant).

However in order there are several changes that I would like to suggest, that would make a conceptual sense:

- Currently the points are not requested in order (sets aren't sorted), we should use a sorted container.
- The loss should be equal to `len(_to_do_seq) / len(sequence)`
- As a corollary, the loss improvement should always be equal to `1/len(sequence)`
- If the original points are numpy arrays, the learner will be requesting tuples, and potentially breaking user-provided functions (e.g. if they call an array method on the point). We should be careful to not change the type of the point.
- We should add a test that the learner correctly requests points in the sequence that was provided by the user.
- The docstring should explain what the learner actually does.

akhmerov:
Actually even easier: loss improvement per point should be just `1`.

basnijholt:
@akhmerov I've implemented your suggestions.

In order to use and pass the exact elements in the sequence that the user passes, I've changed `_to_do_seq` into `_to_do_indices`.

Because `self.pending_points` and self.data` depend on something that is hashable (and the user's input might not be hashable) we use the indices of the input. This means that the user's function receives a tuple `(index, point)` and needs to be wrapped in order to ignore the index.

adaptive/learner/sequence_learner.py
```diff
@@ -0,0 +1,132 @@
...
+    def __setstate__(self, function):
+        self.__init__(function)
+
+
+class SequenceLearner(BaseLearner):
+    r"""A learner that will learn a sequence.
+
+    This is useful when your problem cannot be formulated in terms of
+    another adaptive learner, but you still want to use Adaptive's
+    routines to run, save, and plot.
```
jbweston: I feel there could be a little more explanation about what the learner does. I'm not sure if "learning a sequence" is very clearn.

jbweston: something that explains that the learner simply gives back the points in the provided sequence when `ask`ed.

adaptive/learner/sequence_learner.py
```diff
@@ -0,0 +1,132 @@
+import sys
+from copy import copy
+
+from sortedcontainers import SortedSet, SortedDict
+
+from adaptive.learner.base_learner import BaseLearner
+
+inf = sys.float_info.max
```
jbweston: This isn't needed anymore

PR #195, opened 2019-05-14
don't treat the no data case differently in the Learner1D

basnijholt:


adaptive/learner/learner1D.py
```diff
@@ -590,12 +590,10 @@ def plot(self):
             Plot of the evaluated data.
         """
         hv = ensure_holoviews()
-        if not self.data:
-            p = hv.Scatter([]) * hv.Path([])
-        elif not self.vdim > 1:
+        if not self.vdim > 1:
```
akhmerov: `if self.vdim <= 1`

akhmerov: Ehm, `== 1`, I guess we don't allow `vdim == 0`?

akhmerov:
Other than the comment above LGTM

adaptive/learner/learner1D.py
basnijholt: I vaguely remember that `vdim` could be 0, but looking at the code I don't see how (and why).

I am a bit hesitant to remove it.

akhmerov: Even then `<=` is easier to read than `not >`.

basnijholt: Right, I amended the commit.

Issue #192, opened 2019-05-08
Suggestion: callback feature for learners (or runners)


emascot:
> I'm plotting a 2D Brillouin zone with C3 symmetry. Is there a way to add a callback so that after each point is calculated, I can "tell" the learner to add the same point and value but rotated?

I would like to add points and values after one has finished.
My idea was something like

```python
def callback(learner, point, value):
    learner.tell(transform(point), value)
    ...
```

basnijholt:
Hi @emascot, thanks for your interest in Adaptive and your suggestion! 😄 

I wonder if we could come up with different applications for adding a callback.

For now, I just recommend that you use the `LearnerND` where you use a `ConvexHull` as the bounds, then, when it's finished you can easily do this transformation yourself by looping over the data in `learner.data`.

emascot:
Thanks @basnijholt. I had the same idea. I haven't been able to come up with other applications.

emascot:
Another application I could use this for is to save the learner after each point instead of using periodic saving.

basnijholt:
Hmm, the thing with the callback is that it should be very fast because you want it to happen before a new point is suggested. Saving could be slow and this means that your free cores are waiting for new points. You could in principle get a similar result with a coroutine that frequently checks for the number of points, then if a new point is added, save it again.

Currently, it is actually possible to add coroutines to the runner, see [this part of the tutorial](https://adaptive.readthedocs.io/en/latest/tutorial/tutorial.advanced-topics.html#adding-coroutines), that is what #196 is discussing.

Adding a callback to the learner wouldn't be an asynchronous operation I think and would be a bit different.

@akhmerov and @jbweston do you remember what Pablo tried to do with his KPM code and the `Learner2D`? IIRC he needed to precalculate something with a bunch of vectors before he could calculate a new point. Could this have been solved by having a learner callback?

emascot:
Thanks @basnijholt, I will do this.

PR #196, opened 2019-05-15
add 'runner.block_until_done()' for the 'AsyncRunner'

basnijholt:
It is not very obvious how to use the AsyncRunner in a script. Using this it is easier.

The advantage of using the `AsyncRunner` vs the `BlockingRunner` is that one can add coroutines to the runner that performs certain tasks, like saving, logging, timing, etc.

### DIFF
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -467,6 +467,8 @@ class AsyncRunner(BaseRunner):
         The overhead in percent of using Adaptive. This includes the
         overhead of the executor. Essentially, this is
         ``100 * (1 - total_elapsed_function_time / self.elapsed_time())``.
+    block_until_done : callable
+        Blocks until the ``runner.goal`` has been reached.
 
     Notes
     -----
@@ -651,6 +653,9 @@ class AsyncRunner(BaseRunner):
 
         self.saving_task = self.ioloop.create_task(_saver())
         return self.saving_task
+
+    def block_until_done(self):
+        self.ioloop.run_until_complete(self.task)
 
 
 # Default runner


akhmerov:
Wouldn't callbacks (#192) be a more logical implementation of saving/logging/timing?

basnijholt:
Some things make more sense to do it at a time interval and not after a point is added.

Besides that, I don't see the connection to this PR.

akhmerov:
I'm trying to see whether this API extension is justified.

adaptive/runner.py
```diff
@@ -467,6 +467,8 @@ class AsyncRunner(BaseRunner):
         The overhead in percent of using Adaptive. This includes the
         overhead of the executor. Essentially, this is
         ``100 * (1 - total_elapsed_function_time / self.elapsed_time())``.
+    block_until_done : callable
```
akhmerov: The name and the description is vague: it is unclear what's being blocked.

The original ioloop method name is one better alternative.

basnijholt: I don't think that `run_until_complete` is a good name neither. The `Runner` will always run until it is complete, however, it won't block the kernel.

akhmerov: Perhaps. `wait`/`wait_for_goal` are some alternatives.

Just for my personal understanding: why do we explicitly list methods and their types here? Are there methods that aren't callable?

basnijholt: The only reason I added it there is because the other were already there.

I don't really get why it's there in the first place though.

stuartthomas25:
I've just started using adaptive and I think this would be a useful addition. I want to block some code but also use live_plot while the learners work.

jbweston:
I can't see a reason for this to be added. Let me explain my reasoning.

The use case for this is when you want to add several coroutines that do things in tandem to the runner, but you're not operating in an environment that already has an ioloop running (i.e. a script). Doing this requires knowledge of the ioloop and how to add coroutines etc. Given this, I cannot see what is gained by hiding the call to `ioloop.run_until_complete` in a method.

jbweston:
We should instead add a section to the documentation where we show an example using AsyncRunner, auxiliary tasks, and `ioloop.run_until_complete`.

jbweston:
AFAICT this is covered in the following section of the tutorial: https://adaptive.readthedocs.io/en/latest/tutorial/tutorial.advanced-topics.html#adding-coroutines

PR #197, opened 2019-05-27
Add option to display a progress bar when loading a BalancingLearner

basnijholt:
This adds an option `with_progress_bar` to the `BalancingLearner.load` method.

Output example
image

I am currently trying to load 370k learners which takes more than 30 min on a slow NFS. It would be really useful to display the progress.

### DIFF
--- a/adaptive/learner/balancing_learner.py
+++ b/adaptive/learner/balancing_learner.py
@@ -377,7 +377,7 @@ class BalancingLearner(BaseLearner):
             for l in self.learners:
                 l.save(fname(l), compress=compress)
 
-    def load(self, fname, compress=True):
+    def load(self, fname, compress=True, with_progress_bar=False):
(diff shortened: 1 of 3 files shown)


akhmerov:
I think it's a bad design to expose this to a library user: a good interface should be simple, and this is a customizable non-essential functionality that also adds an extra dependency.

jbweston:
I agree with Anton; I think that this is a functionality that you could implement independently from the library (given that `load` is very simple)

basnijholt:
Thanks for your feedback, I've added it to [`adaptive-scheduler`](https://adaptive-scheduler.readthedocs.io/en/latest/reference/adaptive_scheduler.utils.html#adaptive_scheduler.utils.load_parallel) where it makes more sense because you are dealing with many learners there.

PR #200, opened 2019-06-20
ensure atomic writes when saving a file

basnijholt:
Right now, if a program crashes in the middle of saving, you lose all your data. This ensures that the old file is first moved, then the new file is saved, and only then the old file is removed.

This is not a hypothetical scenario but happens **every day** for some people that I work with ATM.

### DIFF
--- a/adaptive/utils.py
+++ b/adaptive/utils.py
@@ -7,6 +7,8 @@ import pickle
 from contextlib import contextmanager
 from itertools import product
 
+from atomicwrites import AtomicWriter
+
 
 def named_product(**items):
     names = items.keys()
@@ -44,9 +46,13 @@ def save(fname, data, compress=True):
     dirname = os.path.dirname(fname)
     if dirname:
         os.makedirs(dirname, exist_ok=True)
-    _open = gzip.open if compress else open
-    with _open(fname, "wb") as f:
-        pickle.dump(data, f, protocol=pickle.HIGHEST_PROTOCOL)
+
+    blob = pickle.dumps(data, protocol=pickle.HIGHEST_PROTOCOL)
+    if compress:
+        blob = gzip.compress(blob)
+
+    with AtomicWriter(fname, "wb", overwrite=True).open() as f:
+        f.write(blob)
 
 
 def load(fname, compress=True):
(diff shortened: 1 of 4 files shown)


basnijholt:
Maybe adding a `"."` isn't the best name.

adaptive/learner/base_learner.py
```diff
@@ -187,6 +196,12 @@ def load(self, fname, compress=True):
             If the data is compressed when saved, one must load it
             with compression too.
         """
+        with suppress(FileNotFoundError):
```
basnijholt: It's probably better to do this when the loading fails.

akhmerov:
> This is not a hypothetical scenario but happens every day for some people that I work with ATM.

Maybe that is also worth investigating as a separate issue? Saving usually shouldn't cause a problem if adaptive is stable.

akhmerov:
The implementation is rather fragile and potentially hitting bad corner cases (overwriting a different file, times being messed up, etc); I think there should be a more systematic way to do an atomic write.

basnijholt:
> Maybe that is also worth investigating as a separate issue? Saving usually shouldn't cause a problem if adaptive is stable.

The reason for the crash is not Adaptive, but merely low priority nodes that can be evicted at any time, that suddenly stop.

basnijholt:
> The implementation is rather fragile and potentially hitting bad corner cases (overwriting a different file, times being messed up, etc); I think there should be a more systematic way to do an atomic write.

Do you have a suggestion?

akhmerov:
https://github.com/untitaker/python-atomicwrites seems like an option or any other alternatives.

Or we could use all the relevant tricks from it.

akhmerov:
I also have to bring up another approach: we provide a way to get an in-memory data in adaptive and the user uses `atomicwrites` on their own. Separation of concerns and all.

basnijholt:
> I also have to bring up another approach: we provide a way to get an in-memory data in adaptive and the user uses atomicwrites on their own. Separation of concerns and all.

There is already [`_get_data` and `_set_data`](https://github.com/python-adaptive/adaptive/blob/master/adaptive/learner/base_learner.py#L148) that every learner implements. The problem is that you can't use the `periodic_saver` like this

I do not mind to add the dependency, it's used by 19723 other packages and it is tiny. See the commit I've made. Also, I don't see a reason why people wouldn't want a 'safe' save.

PR #201, opened 2019-06-24
Update tutorial.parallelism.rst

aeantipov:

PR #202, opened 2019-06-24
Authors

basnijholt:
https://github.com/python-adaptive/adaptive/pull/201 failed because of a missing author.

This PR allows the author test to fail and separates current maintainers from contributors.

`ContinueOnError: true` [is the same as WarnAndContinue](https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-ignore-errors-in-tasks?view=vs-2019#use-the-continueonerror-attribute).

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -1,8 +1,13 @@
 ## Authors
-Below is a list of the contributors to Adaptive:
 
-+ [Anton Akhmerov](<https://antonakhmerov.org>)
+The current maintainers of Adaptive are:
+
 + [Bas Nijholt](<http://nijho.lt>)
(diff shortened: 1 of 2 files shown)


akhmerov:
I believe, authors check should be a part of the linting pipeline. Let's move it there.

basnijholt:
It already is in the linting pipeline.

akhmerov:
@jbweston what's your opinion specifically on the proposed reorganization of the authors file? (that's the main decision to be made here)

jbweston:
> @jbweston what's your opinion specifically on the proposed reorganization of the authors file? (that's the main decision to be made here)

I'm fine with the Authors file reorg

PR #204, opened 2019-07-10
use jupyter_sphinx v0.2.0 from conda instead of my branch

basnijholt:
This again needs https://github.com/python-adaptive/adaptive/pull/169. However, I have manually executed those commands from the CI.

See https://adaptive.readthedocs.io/en/test_documentation/ when the build finished.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -22,7 +22,7 @@ import adaptive  # noqa: E402
 # -- Project information -----------------------------------------------------
 
 project = "adaptive"
-copyright = "2018, Adaptive Authors"
+copyright = "2018-2019, Adaptive Authors"
 author = "Adaptive Authors"
 
 # The short X.Y version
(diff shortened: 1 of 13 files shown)


akhmerov:
Let's pin jupyter-sphinx at `<= 0.3`, and specify the thebe config like [here](https://jupyter-sphinx.readthedocs.io/en/latest/#thebelab-support). Pointing to this repo should be all it takes.

basnijholt:
Actually, I rather pin it to a fixed version like everything else.

Is adding those lines to the config really enough to install thebelab?

Of course, it would need some buttons.

docs/source/conf.py
```diff
@@ -83,6 +83,12 @@
 pygments_style = "sphinx"
 
 
+jupyter_sphinx_thebelab_config = {
+    "requestKernel": True,
+    "binderOptions": {"repo": "binder-examples/requirements"},
```
akhmerov: `  "binderOptions": {"repo": "python-adaptive/adaptive"},`

basnijholt:
Also, how would we make sure that this gets executed https://github.com/python-adaptive/adaptive/blame/master/docs/source/tutorial/tutorial.Learner1D.rst#L13-L21

akhmerov:
> Also, how would we make sure that this gets executed https://github.com/python-adaptive/adaptive/blame/master/docs/source/tutorial/tutorial.Learner1D.rst#L13-L21

These will become visible but the user will need to manually execute those.

jbweston:
I updated the dependency to jupyter_sphinx 0.2.1. All the other dependencies are pinned at exact version, so I think it's fine

jbweston:
This LGTM

PR #209, opened 2019-08-26
remove MPI4PY_MAX_WORKERS where it's not used

basnijholt:


akhmerov:
I have no experience with MPI, so I cannot check if this code is correct, but if you know—go ahead and merge.

PR #210, opened 2019-08-26
WIP: raise an error when using a lambda and default executor

### LINKED ISSUE #206: Does not work with lambda functions
For example, if in the initial example one have
```
def g(x, offset=offset, wait=True):
    from time import sleep
    from random import random

    a = 0.01
    if wait:
        sleep(random() / 10)
    return x + a**2 / (a**2 + (x - offset)**2)


f = lambda x: g(x)
```
then execution just hangs.

basnijholt:
Closes #206.

It is not obvious that this blocks the kernel, but it does hang forever and does not error.

```python
import adaptive
adaptive.notebook_extension()

f = lambda xy: xy[0]**2 + xy[1]**2
learner = adaptive.Learner2D(f, bounds=[(-1, 1), (-1, 1)])
runner = adaptive.Runner(learner, goal=lambda l: l.npoints >= 1000)
```

This PR makes sure an error is raised, however, might this be caused by an underlying issue?

### DIFF
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -8,6 +8,7 @@ import os
 import sys
 import time
 import traceback
+import types
 import warnings
 from contextlib import suppress
 
@@ -493,6 +494,14 @@ class AsyncRunner(BaseRunner):
 
             def goal(_):
                 return False
+
+        if isinstance(learner.function, types.LambdaType) and executor is None:
+            raise ValueError(
+                "A lambda function cannot be pickled and "
+                "therefore doesn't work with the default executor."
+                "Either do not use a lamdba or use a framework that"
+                " allows this, i.e. `ipyparallel` with `dill`."
+            )
 
         super().__init__(
             learner,
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -5,6 +5,7 @@ import asyncio
 import concurrent.futures as concurrent
 import inspect
 import os
+import pickle
 import sys
 import time
 import traceback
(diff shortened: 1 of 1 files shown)


akhmerov:
Here's my thinking:

- #206 is likely not a pressing issue
- There are other other common cases where pickling fails (e.g. closures). Those aren't going to be caught by a type check.

Therefore I propose to not apply a patch like this, but rather wait until the issue becomes more important, or someone (maybe one of us) is motivated to have a proper fix.

basnijholt:
To be fair, two persons now reported this issue separately.

See https://github.com/python-adaptive/adaptive/issues/211 for the other one.

akhmerov:
Perhaps a better implementation would be to try to pickle the function on our own and see if that raises?

basnijholt:
I have added that check.

I still also explicitly check whether it is a lambda because that is the most common case I think.

adaptive/runner.py
```diff
@@ -494,6 +496,19 @@ def __init__(
             def goal(_):
                 return False
 
+        if executor is None:
+            if isinstance(learner.function, types.LambdaType):
+                raise ValueError(
+                    "A lambda function cannot be pickled and "
+                    "therefore doesn't work with the default executor."
+                    "Either do not use a lamdba or use a framework that"
```
jbweston:
```suggestion
                    "Either do not use a lamdba or use an executor that"
```

jbweston: Maybe we should give the specific example as a code snippet in the error message instead of just saying " `ipyparallel` with `dill`"

adaptive/runner.py
```diff
@@ -494,6 +496,19 @@ def __init__(
             def goal(_):
                 return False
 
+        if executor is None:
+            if isinstance(learner.function, types.LambdaType):
```
jbweston: this check doesn't actually check if the function is a `lambda`. Check it out:

```
import types

def f():
    pass

print(isinstance(f, types.LambdaType))
```

prints `True`

jbweston: This is why the tests are failing AFAICT

basnijholt: I've made the check a bit simpler.

basnijholt:
I actually ran into this issue myself today!

I also simplified the test a bit. @jbweston or @akhmerov should I rephrase the error message or merge as is?

basnijholt:
Kim just ran into this problem too.

jbweston:
If we fix the test failures we can merge!

Issue #88, opened 2018-12-19
Introduce "Learner not ready" error


basnijholt:
## ([original issue on GitLab](https://gitlab.kwant-project.org/qt/adaptive/issues/51))

_opened by Anton Akhmerov ([@anton-akhmerov](https://gitlab.kwant-project.org/anton-akhmerov)) at 2018-03-10T20:58:36.245Z_

Some adaptive algorithms may be implemented such that they are not ready to request new data until more data arrives. While this reduces the suitability of these algorithms to parallelization, there is no reason to prohibit it.

We could handle this as follows:
* Introduce an `InsufficientData` error (name is tentative) that could be a subclass of `StopIteration` if we view learners as iterators (see [gitlab:#45](https://gitlab.kwant-project.org/qt/adaptive/issues/45)).
* Teach the runners and `BalancingLearner` to catch this error and only resume querying the learner after they supply additional data.

That way we could even write a trivial wrapper learner that works with any adaptive algorithm and blocks on every step.

basnijholt:
_originally posted by Christoph Groth ([@cwg](https://gitlab.kwant-project.org/cwg)) at 2018-04-20T12:17:06.858Z on [GitLab](https://gitlab.kwant-project.org/qt/adaptive/issues/51)_

Wouldn't returning `None` when new points are requested be more natural?  `StopIteration` has the meaning that iteration over this iterator is finished, so your proposal would require getting a new iterator from the learner once data has been fed back.  That seems strange.

Consider evaluating the same function with the same deterministic learner but with different resources.  The result of the learning will be the same in all cases.  It seems strange to couple the lifetime of iterators to the availability of resources.

To me, a more natural use of StopIteration is to allow the learner to signal the runner when the a precision condition is met (when specified, the default should be infinite iteration).

While it can be useful to have custom conditions or even run the learner until it is stopped manually, very often some criterion of precision can be specified upfront.  (Think of adaptive integration which is currently the most commonly used case of adaptive evaluation.)  I have the impression that the learner is the best place to evaluate whether such criteria are reached or not.

basnijholt:
_originally posted by Anton Akhmerov ([@anton-akhmerov](https://gitlab.kwant-project.org/anton-akhmerov)) at 2018-04-20T12:19:20.860Z on [GitLab](https://gitlab.kwant-project.org/qt/adaptive/issues/51)_

You're right, returning `None` is a more intuitive interface for indicating that the learner isn't ready.

PR #214, opened 2019-09-04
fix problem when learner has no more points for BalancingLearner

basnijholt:


adaptive/learner/balancing_learner.py
```diff
@@ -119,20 +119,27 @@ def strategy(self, strategy):
...
     def _ask_and_tell_based_on_loss_improvements(self, n):
         selected = []  # tuples ((learner_index, point), loss_improvement)
         total_points = [l.npoints + len(l.pending_points) for l in self.learners]
         for _ in range(n):
-            to_select = []
-            for index, learner in enumerate(self.learners):
-                # Take the points from the cache
-                if index not in self._ask_cache:
-                    self._ask_cache[index] = learner.ask(n=1, tell_pending=False)
-                points, loss_improvements = self._ask_cache[index]
```
basnijholt: here one would need a double break, but because that doesn't exist I make it into a function.

basnijholt:
Wow! This seems to break a lot of tests, will investigate some time later.

adaptive/learner/balancing_learner.py
```diff
@@ -119,27 +119,34 @@ def strategy(self, strategy):
...
 
+    def _ask_all_learners(self, total_points):
+        to_select = []
+        for index, learner in enumerate(self.learners):
+            # Take the points from the cache
+            if index not in self._ask_cache:
+                self._ask_cache[index] = learner.ask(n=1, tell_pending=False)
+            points, loss_improvements = self._ask_cache[index]
+            if not points:  # cannot ask for more points
+                return to_select
```
jbweston: There are a couple of things I don't understand here:

+ seems this `return` should be a `continue`; just because learner `i` could not give any more points does not mean that no other learners can give any!
+ I cannot see when this branch will ever be executed. `learner.ask(1)` is guaranteed to return a point. At the moment there is no way for a learner to indicate that it has "no more points". If a learner returns no points then it is in violation of the API and other stuff is liable to break

akhmerov:
> It may indeed be a good idea for learners to be able to indicate that they have no more points, but this is something that needs to be looked at carefully and should not just be implemented in an ad-hoc way.

Indeed, this is #88

jbweston:
Seems to me that this should be closed, and we should refer to #88

PR #215, opened 2019-09-05
add scatter_or_line argument to Learner1D.plot

basnijholt:
I've needed this many times and the discussion in the chat now also made it clear that this is needed.

Now you won't have to look at plots like this: 
image

### DIFF
--- a/adaptive/learner/learner1D.py
+++ b/adaptive/learner/learner1D.py
@@ -580,7 +580,7 @@ class Learner1D(BaseLearner):
         loss = mapping[ival]
         return finite_loss(ival, loss, self._scale[0])
 
-    def plot(self):
+    def plot(self, *, scatter_or_line=None):
         """Returns a plot of the evaluated data.
 
         Returns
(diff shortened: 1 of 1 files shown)


akhmerov:
Not sure; we're already using not-a-plotting-library (holoviews), so I'd rather keep adaptive's plotting interface as small as possible.

adaptive/learner/learner1D.py
```diff
@@ -580,22 +580,33 @@ def _loss(self, mapping, ival):
...
-    def plot(self):
+    def plot(self, *, scatter_or_line=None):
         """Returns a plot of the evaluated data.
 
         Returns
         -------
         plot : `holoviews.element.Scatter` (if vdim=1)\
                else `holoviews.element.Path`
             Plot of the evaluated data.
+        scatter_or_line : str, optional
```
jbweston: This is a parameter, not a return value

jbweston: also set the default value to `scatter` and mention this in the docstring

adaptive/learner/learner1D.py
```diff
@@ -580,22 +580,33 @@ def _loss(self, mapping, ival):
...
+            Plot as a scatter plot ("scatter") or a line plot ("line").
+            By default a line plot will be chosen if the data consists of
+            vectors, otherwise it is a scatter plot.
         """
         hv = ensure_holoviews()
 
         xs, ys = zip(*sorted(self.data.items())) if self.data else ([], [])
         if self.vdim == 1:
-            p = hv.Path([]) * hv.Scatter((xs, ys))
+            if scatter_or_line is None or scatter_or_line == "scatter":
```
jbweston: once `scatter_or_line` has a default value of `scatter` you can change this test and the one below to be `if scatter_or_line == "scatter`

adaptive/learner/learner1D.py
```diff
@@ -580,22 +580,33 @@ def _loss(self, mapping, ival):
...
         -------
         plot : `holoviews.element.Scatter` (if vdim=1)\
                else `holoviews.element.Path`
             Plot of the evaluated data.
+        scatter_or_line : str, optional
+            Plot as a scatter plot ("scatter") or a line plot ("line").
+            By default a line plot will be chosen if the data consists of
+            vectors, otherwise it is a scatter plot.
         """
         hv = ensure_holoviews()
```
jbweston: add a check here:
```
if scatter_or_line not in ("scatter", "line"):
    raise ValueError("scatter_or_line must be 'scatter' or 'line'")
```

basnijholt: Done!

adaptive/learner/learner1D.py
basnijholt: Simplified it a bit more.

adaptive/learner/learner1D.py
basnijholt: Done!

PR #216, opened 2019-09-10
2D: add interpolated_on_grid method

basnijholt:
@michaelwimmer told me this would be a useful addition. He (and many others before) struggled to get the interpolated data from the learner.

@michaelwimmer what do you think of this?

### DIFF
--- a/adaptive/learner/learner2D.py
+++ b/adaptive/learner/learner2D.py
@@ -476,7 +476,7 @@ class Learner2D(BaseLearner):
             points, values = self._data_in_bounds()
             return interpolate.LinearNDInterpolator(points, values)
 
-    def _interpolate_combined(self):
+    def _interpolator_combined(self):
         """A `scipy.interpolate.LinearNDInterpolator` instance
         containing the learner's data *and* interpolated data of
         the `pending_points`."""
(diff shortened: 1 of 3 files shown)


akhmerov:
Just a quick feedback. I think it's a useful convenience method, well spotted. We should probably also use it internally in plotting.

I am on the other hand unsure about the interface choices (method name, input parameters, return values).

basnijholt:
It is already used inside the plotting method.

I think it’s reasonable. What do you suggest?

akhmerov:
Hm, what about `learner.interpolate(x, y)`, where `x` and `y` would use numpy broadcasting rules (so if x is N×1 and y is `1×N`, then all combinations are evaluated.

That would still require the user to provide the limits, so we could also use `None` as a default for both, and that would use a logic similar to the current one.

michaelwimmer:
@akhmerov there is also ip() and ip_combined() - don't these do a bit what you suggest? 

It is also useful to get the underlying grid where the interpolated values live on, because this is not totally trivial (need to make sure you do not get out of the valid region of the interpolation) and is something most people need (and `np.meshgrid` and `np.mgrid` are hard)

jbweston:
@anton @basnijholt points out that providing `x` and `y` allows people to provide data outside the interpolated area, which the interpolator can't handle. This also includes the common case where someone uses one of the boundary points for a limit.

basnijholt:
@akhmerov and @jbweston, I've renamed `data_on_grid` to `interpolated_on_grid`.

Also, I have renamed the vaguely named `ip` to ``interpolator` and added an argument `scaled` + example.

akhmerov:
LGTM

PR #219, opened 2019-09-20
pass value_scale to the LearnerND's loss_per_simplex function

basnijholt:
Currently, the `LearnerND` cannot access the unscaled value in the `loss_per_simplex`.

This PR fixes this by passing the scale. This fixes a regression introduced when we added scaling.

Now the following code:
```python
import numpy as np
import adaptive
adaptive.notebook_extension()

def isoline_loss_function(y_iso, sigma, priority=1):
    from adaptive.learner.learnerND import default_loss

    def gaussian(x, mu, sigma):
        return np.exp(-(x - mu) ** 2 / sigma ** 2)

    def loss(simplex, values, value_scale):
        distance = np.mean([abs(y_iso * value_scale - y) for y in values])
        return priority * gaussian(distance, 0, sigma) + default_loss(simplex, values, value_scale)

    return loss

level = 0.5
loss = isoline_loss_function(y_iso=level, sigma=1, priority=0.3)

def f(xy):
    x, y = xy
    return x ** 2 + y ** 2 / 2 + y ** 3

learner = adaptive.LearnerND(
    f,
    bounds=[(-1, 1), (-1, 1)],
    loss_per_simplex=loss,
)

adaptive.runner.simple(learner, goal=lambda l: l.npoints > 1000)
%opts Image [colorbar=True]
learner.plot_isoline(level, tri_alpha=0.3)
```

produces
image

instead of 

image

### DIFF
--- a/adaptive/learner/learnerND.py
+++ b/adaptive/learner/learnerND.py
@@ -47,12 +47,45 @@ def orientation(simplex):
     return sign
 
 
-def uniform_loss(simplex, ys=None):
+def uniform_loss(simplex, values, value_scale):
+    """
+    Uniform loss.
+
+    Parameters
+    ----------
+    simplex : list of tuples
+        Each entry is one point of the simplex.
+    values : list of values
+        The scaled function values of each of the simplex points.
+    value_scale : float
+        The scale of values, where ``values = function_values * value_scale``.
+
+    Returns
+    -------
+    loss : float
+    """
     return volume(simplex)
 
 
-def std_loss(simplex, ys):
-    r = np.linalg.norm(np.std(ys, axis=0))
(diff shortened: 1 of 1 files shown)


jbweston:
Seems ok, don't know why tests are failing

jbweston:
@basnijholt the test failures are genuine. 

```
>      loss_curvature = triangle_loss(simplex, values, neighbors, neighbor_values)
E       TypeError: triangle_loss() missing 1 required positional argument: 'neighbor_values'
```

Threre may be others

basnijholt:
Thanks! I've already pushed a fix.

basnijholt:
@jbweston merge if you are OK with this 👍

PR #220, opened 2019-10-06
WIP: unify Learner1D, Learner2D and LearnerND

jbweston:
While writing the adaptive paper we managed to write down a simple algorithm formulated in terms of abstract domains and subdomains.

Implementing such an algorithm in a learner should allow us to unify Learner1D, Learner2D and LearnerND.


jbweston:
The following are not yet supported:
+ 2D and 3D domains
+ BalancingLearner (requires removing points from the interior of domains)

akhmerov:
Also compared to the existing implementation, global rescaling is missing. (e.g. computing y-scale of the values, and normalizing the data by it)

jbweston:
> Also compared to the existing implementation, global rescaling is missing. (e.g. computing y-scale of the values, and normalizing the data by it)

Should this perhaps be something that the loss function does itself?

akhmerov:
> I could imagine extending the loss function signature to include the x and y scales.

Indeed, but then the learner needs two extra hooks: one at each step to update global metrics, and another one to trigger loss recomputation for all subdomains once the global metrics changes sufficiently much so that old losses become obsolete.

jbweston:
> Indeed, but then the learner needs two extra hooks: one at each step to update global metrics, and another one to trigger loss recomputation for all subdomains once the global metrics changes sufficiently much so that old losses become obsolete.

I added this now

adaptive/learner/new_learnerND.py
```diff
@@ -0,0 +1,513 @@
...
+
+        self.queue = Queue()
+        self.data = dict()
+        self.function = f
+
+        # Evaluate boundary points right away to avoid handling edge
+        # cases in the ask and tell logic
+        bound_points = sorted(map(tuple, itertools.product(*bounds)))
+        for x in bound_points:
+            y = f(x)
```
basnijholt: Does this really need to happen in `__init__`?

jbweston: It simplifies everything else if we do. We can lift this restriction at the end, but I'd rather not complicate the implementation of `ask` for the moment.

jbweston:
### TODO

+ [ ] don't evaluate boundary points in `__init__`
+ [ ] revisit the loss function signature
+ [ ] add tests

basnijholt:
All other learners implement `pending_points` which is a set.

Would that change anything? Now I see you set `self.data[x] = None`.

jbweston:
>All other learners implement pending_points which is a set. Would that change anything? Now I see you set self.data[x] = None.

I'm using `pending_points` now

adaptive/learner/new_learnerND.py
```diff
@@ -758,27 +166,30 @@ def _finalize_initialization(self):
...
 
     def ask(self, n, tell_pending=True):
         if self.n_asked >= len(self.boundary_points):
             points, losses = self._ask(n, tell_pending)
         else:
-            points = self.boundary_points[self.n_asked:self.n_asked + n]
+            points = self.boundary_points[self.n_asked : self.n_asked + n]
             # The boundary points should always be evaluated with the highest priority
-            losses = [float('inf')] * len(points)
+            losses = [float("inf")] * len(points)
```
akhmerov: The [canonical place](https://docs.python.org/3/library/math.html#math.inf) for infinity is `math.inf`.

adaptive/priority_queue.py
```diff
@@ -0,0 +1,104 @@
...
+        # items may not be hashable we cannot use a SortedDict, so we use a
+        # SortedList storing '(item, key)'.
+        self._items = SortedList(((v, k) for k, v in self._queue.items()))
+
+    def __len__(self):
+        return len(self._queue)
+
+    def items(self):
+        "Return an iterator over the items in the queue in arbitrary order."
+        return reversed(self._queue.values())
```
akhmerov: Why do we use `reversed` here and later if the order is arbitrary?

jbweston: changed the API to return in priority order

adaptive/learner/new_learnerND.py
jbweston: fixed

jbweston:
Where should we put the new LearnerND? Or maybe we should call it something different

basnijholt:
I say, overwrite the other LearnerND.

jbweston:
I added the new learner to all the learner tests except the following

+ `test_uniform_sampling`: this test is marked `xfail` anyway
+ `test_point_adding_order_is_irrelevant`: This is marked xfail for Learner2D and LearnerND anyway, and I'm not sure if this behaviour should be satisfied. If we add points in a different order I think we can in principle end up with different triangulations (?)

The new learner is marked xfail on the following tests:

+ `test_learner_performance_is_invariant_under_scaling`: Currently none of the loss functions scale the points (because I was lazy when implementing) but they probably should

jbweston:
Now I belive the only 2 things left to do are:

+ [ ] decide on the loss function signature (at the moment the loss function gets all data etc.
+ [ ] implement scaling within the loss functions

jbweston:
There's also the question of how to review this MR. It's a lot of code.

It may also be that this implementation is inferior to the current LearnerND. For example, the old LearnerND is 1180 lines, whereas this new implementation in 1350 (including the priority queue, domain definitions etc0

adaptive/learner/triangulation.py
```diff
@@ -266,7 +266,7 @@ class Triangulation:
         or more simplices in the
     """
 
-    def __init__(self, coords):
+    def __init__(self, coords, *, _check_vertices=True):
```
akhmerov: Do we need an underscore if this is already a keyword-only argument?

adaptive/learner/triangulation.py
```diff
@@ -287,23 +287,28 @@ def __init__(self, coords):
...
-        if np.linalg.matrix_rank(vectors) < dim:
-            raise ValueError(
-                "Initial simplex has zero volumes "
-                "(the points are linearly dependent)"
-            )
+        if _check_vertices:
+            vectors = np.subtract(coords[1:], coords[0])
+            if np.linalg.matrix_rank(vectors) < dim:
+                raise ValueError(
+                    "Initial simplex has zero volumes "
```
akhmerov: It's not necessarily a single simplex at this point; this should read "Hull has a zero volume", right?

jbweston: Yes. This error message was already wrong before I touched it. Maybe I should update it

adaptive/learner/triangulation.py
jbweston: I meant to signal that you should only use this if you know what you're doing; I didn't even add it to the docstring!

basnijholt: I think it needs both. It should definitely not be a able to use as a positional keyword by accident nor should it be changed by the user.

adaptive/domain.py
```diff
@@ -0,0 +1,743 @@
...
+import scipy.linalg
+import scipy.spatial
+from sortedcontainers import SortedList
+
+from adaptive.learner.triangulation import Triangulation, circumsphere, point_in_simplex
+
+__all__ = ["Domain", "Interval", "ConvexHull"]
+
+
+class Domain(metaclass=abc.ABCMeta):
```
akhmerov: I think it would be useful to add a class docstring with a description of the important properties of this object.

adaptive/domain.py
```diff
@@ -0,0 +1,743 @@
...
+
+    def __init__(self, a, b):
+        if a >= b:
+            raise ValueError("'a' must be less than 'b'")
+
+        # If a sub-interval contains points in its interior, they are stored
+        # in 'sub_intervals' in a SortedList.
+        self.bounds = (a, b)
+        self.sub_intervals = dict()
+        self.points = SortedList([a, b])
```
akhmerov: It would in principle be sufficient to store only the boundaries of the interval and always query for pairs of neighboring points. The current implementation is quite OK OTOH.

PR #221, opened 2019-10-09
2D: add triangle_loss

basnijholt:
For symmetry reasons, I add `triangle_loss` which works with the `Learner2D`.

Currently, it is very slow, for a learner with 1000 points:
```python
ip = learner.ip()
%timeit adaptive.learner.learner2D.triangle_loss(ip)
%timeit adaptive.learner.learner2D.default_loss(ip)
```
```
652 ms ± 154 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
1.92 ms ± 96.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Result:
image

### DIFF
--- a/adaptive/learner/learner2D.py
+++ b/adaptive/learner/learner2D.py
@@ -9,6 +9,7 @@ import numpy as np
 from scipy import interpolate
 
 from adaptive.learner.base_learner import BaseLearner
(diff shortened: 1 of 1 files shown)


basnijholt:
@pulkin has written some C code [here](https://github.com/pulkin/curious) that might speed up this calculation.

PR #222, opened 2019-10-14
add _RequireAttrsABCMeta and make the BaseLearner use it

basnijholt:
This makes sure that the Learners will conform to the default API.

### DIFF
--- a/adaptive/learner/base_learner.py
+++ b/adaptive/learner/base_learner.py
@@ -4,7 +4,7 @@ import abc
 from contextlib import suppress
 from copy import deepcopy
 
-from adaptive.utils import load, save
+from adaptive.utils import load, save, _RequireAttrsABCMeta
 
 
 def uses_nth_neighbors(n):
@@ -61,7 +61,7 @@ def uses_nth_neighbors(n):
     return _wrapped
 
 
-class BaseLearner(metaclass=abc.ABCMeta):
+class BaseLearner(metaclass=_RequireAttrsABCMeta):
     """Base class for algorithms for learning a function 'f: X → Y'.
 
     Attributes
@@ -84,6 +84,8 @@ class BaseLearner(metaclass=abc.ABCMeta):
     Subclasses may define a ``plot`` method that takes no parameters
     and returns a holoviews plot.
     """
+
+    required_attributes = ["data", "npoints", "pending_points"]
 
     def tell(self, x, y):
         """Tell the learner about a single value.
--- a/adaptive/utils.py
+++ b/adaptive/utils.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 
+import abc
 import functools
 import gzip
 import os
@@ -67,3 +68,14 @@ def copy_docstring_from(other):
         return functools.wraps(other)(method)
 
     return decorator
+
+
+class _RequireAttrsABCMeta(abc.ABCMeta):
+    required_attributes = []
+
+    def __call__(self, *args, **kwargs):
+        obj = super().__call__(*args, **kwargs)
+        for name in obj.required_attributes:
+            if not hasattr(obj, name):
+                raise ValueError(f"Required attribute {name} not set in __init__.")
+        return obj


adaptive/utils.py
```diff
@@ -67,3 +68,14 @@ def decorator(method):
...
+
+
+class _RequireAttrsABCMeta(abc.ABCMeta):
+    required_attributes = []
+
+    def __call__(self, *args, **kwargs):
+        obj = super().__call__(*args, **kwargs)
+        for name in obj.required_attributes:
+            if not hasattr(obj, name):
+                raise ValueError(f"Required attribute {name} not set in __init__.")
```
jbweston: We should also check that the type is good.

We should use class attribute annotations to specify this:

```
class A:
    x: int
    y: float
```

adaptive/learner/base_learner.py
```diff
@@ -85,6 +85,8 @@ class BaseLearner(metaclass=abc.ABCMeta):
     and returns a holoviews plot.
     """
 
+    required_attributes = ["data", "npoints", "pending_points"]
+
```
jbweston: change this to class annotations (see below)

basnijholt:
@akhmerov and @jbweston, I've implemented this as we discussed and fixed the issues in the learners that didn't implement these attributes.

adaptive/utils.py
```diff
@@ -67,3 +68,18 @@ def decorator(method):
...
+        obj = super().__call__(*args, **kwargs)
+        for name, type_ in obj.__annotations__.items():
+            try:
+                x = getattr(obj, name)
+                if not isinstance(x, type_):
+                    raise TypeError(
+                        f"The attribute '{name}' is of {type_} instead of {type(x)}."
+                    )
+            except AttributeError as e:
+                raise e(f"Required attribute {name} not set in __init__.")
```
jbweston:
```
            try:
                x = getattr(obj, name)
            except AttributeError:
                raise AttributeError(f"Required attribute {name} not set in __init__.") from None
            else:
                if not isinstance(x, type_):
                    raise TypeError(
                        f"The attribute '{name}' is of {type_} instead of {type(x)}."
                    )
```

jbweston: your
```
except AttributeError as e:
    raise e("...")
```
will not work; you can't call an instance of an exception like a function

adaptive/learner/base_learner.py
```diff
@@ -61,30 +61,31 @@ def _wrapped(loss_per_interval):
...
-class BaseLearner(metaclass=abc.ABCMeta):
+class BaseLearner(metaclass=_RequireAttrsABCMeta):
     """Base class for algorithms for learning a function 'f: X → Y'.
 
     Attributes
     ----------
     function : callable: X → Y
-        The function to learn.
+        The function to learn, the ``learner`` might modify
+        the original function.
```
jbweston: It's not really clear what "the original function" means here. The base class doesn't have an `__init__` or anything

adaptive/utils.py
basnijholt: Oops...

adaptive/utils.py
```diff
@@ -67,3 +68,18 @@ def decorator(method):
...
+        obj = super().__call__(*args, **kwargs)
+        for name, type_ in obj.__annotations__.items():
+            try:
+                x = getattr(obj, name)
+                if not isinstance(x, type_):
+                    raise TypeError(
+                        f"The attribute '{name}' is of {type_} instead of {type(x)}."
+                    )
+            except AttributeError:
+                raise AttributeError(f"Required attribute {name} not set in __init__.")
```
jbweston: You didn't implement the changes I suggested.
There are still 2 things wrong with this:
+ the `try` block extends over too much code: you actually want it to extend only over the `x = getattr(obj, name)`
+ the `raise` in the `except` block will give spurious backtrace unless you `raise ... from None` like I suggested.

Once again:


```
try:
    x = getattr(obj, name)
except AttributeError:
    raise AttributeError(f"Required attribute {name} not set in __init__.") from None
else:
    if not isinstance(x, type_):
        raise TypeError(
            f"The attribute '{name}' is of {type_} instead of {type(x)}."
        )
```

jbweston: Also the error message for the TypeError was confusing IMO; I suggest an alternative for that above

basnijholt: Oh, sorry, I didn't see https://github.com/python-adaptive/adaptive/pull/222#discussion_r335838905 before.

I've changed the code into what you suggested.

adaptive/learner/base_learner.py
basnijholt: I've changed it. Better?

adaptive/utils.py
```diff
@@ -67,3 +68,20 @@ def decorator(method):
...
+
+class _RequireAttrsABCMeta(abc.ABCMeta):
+    def __call__(self, *args, **kwargs):
+        obj = super().__call__(*args, **kwargs)
+        for name, type_ in obj.__annotations__.items():
+            try:
+                x = getattr(obj, name)
+            except AttributeError:
+                raise AttributeError(
+                    f"Required attribute {name} not set in __init__."
```
akhmerov: What is clearer: "not set in __init__", "missing in the instance", "not set during instantiation"?

PR #230, opened 2019-11-28
WIP: replace BlockingRunner with AsyncRunner + asyncio.run

jbweston:
This eliminates a lot of code duplication.
We also deprecate the use of BlockingRunner directly.

### DIFF
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -287,8 +287,28 @@ class BaseRunner(metaclass=abc.ABCMeta):
         pass
 
 
-class BlockingRunner(BaseRunner):
+def BlockingRunner(
+    learner,
+    goal,
+    *,
+    executor=None,
+    ntasks=None,
+    log=False,
+    shutdown_executor=False,
+    retries=0,
+    raise_if_retries_exceeded=True,
+):
     """Run a learner synchronously in an executor.
+
+    **This runner is deprecated**, replace:
+
+    >>> r = adaptive.BlockingRunner(...)
+
+    with
+
+    >>> import asyncio
+    >>> r = adaptive.Runner(...)
+    >>> asyncio.run(r)
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


adaptive/runner.py
```diff
@@ -347,62 +367,30 @@ class BlockingRunner(BaseRunner):
...
+        ntasks=ntasks,
+        log=log,
+        shutdown_executor=shutdown_executor,
+        retries=retries,
+        raise_if_retries_exceeded=raise_if_retries_exceeded,
+    )
 
-    def _run(self):
-        first_completed = concurrent.FIRST_COMPLETED
+    asyncio.run(r)
```
basnijholt: This is Python 3.7 only :(

jbweston:
## TODO

+ [ ] use the [deprecated](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) sphinx directive in docstrings
+ [ ] replaces uses of

adaptive/runner.py
basnijholt: but `self.ioloop.run_until_complete(r)` should do the same.

jbweston: We could also bump the version requirement

basnijholt: Seems good to me!

adaptive/runner.py
```diff
@@ -287,9 +287,29 @@ def _submit(self, x):
...
 
+    **This runner is deprecated**, replace:
+
+    >>> r = adaptive.BlockingRunner(...)
+
+    with
+
+    >>> import asyncio
+    >>> r = adaptive.Runner(...)
+    >>> asyncio.run(r)
```
jbweston: This is wrong for a couple of reasons:

1. `Runner` is not a coroutine, so we can't use it with `asyncio.run`
2. `r.task` is a *task* and hence tied to a given ioloop.

This really should look like

```suggestion
    >>> asyncio.get_event_loop().run_until_complete(r.task)
```

adaptive/runner.py
jbweston:
```suggestion
    r.ioloop.run_until_complete(r.task)
```

PR #231, opened 2019-11-28
WIP: add isosurface loss

jbweston:
This is the "simplest thing that could work" that I worked on with @akhmerov and @oroszl last week.

Probably by thinking for 5 minutes we could come up with something even better that's not any more complicated to implement.


jbweston:
Another idea that we came up with: make the loss proportional to the *area* that is cut by the isosurface

PR #232, opened 2019-11-28
add a changelog

jbweston:
We move fast and break things. For everyone's sanity it would be good to keep a changelog where we document what stuff we added, fixed and changed.

## TODO
+ [x] Add changelog entries for the previous releases  (using the release notes)
+ [ ] Add changelog entry for the unreleased changes
+ [ ] Add the sphinx [versionadded](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded) directive to relevant docstrings so that we know when stuff was added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,7 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [unreleased]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,209 +1,398 @@
 # Changelog
-All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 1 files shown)


basnijholt:
The changes are documented to some degree [here](https://github.com/python-adaptive/adaptive/releases).

jbweston:
> The changes are documented to some degree [here](https://github.com/python-adaptive/adaptive/releases).

That's where we'll get the information to help with bullet point (1), but it is useful to keep this information in the repository because then we can do things like add it to the documentation

basnijholt:
I've added the entries from https://github.com/python-adaptive/adaptive/releases.

I think we can already merge this and open issues for the other two points @jbweston brought up.

jbweston:
@basnijholt do you mind if I fixup the changelog to separate the issues solved into bugfixes and features, in order to conform to the KeepAChangelog format?

jbweston:
:+1: on making the other 2 things separate issues

basnijholt:
I am going to go ahead and merge this. 

@jbweston, the things you wanted to fix, we can always do later. Having _any_ changelog is more valuable than not having it.

basnijholt:
See https://github.com/python-adaptive/adaptive/issues/306.

Issue #233, opened 2019-12-05
SKOptLearner doesn't work for multi variate domain


caenrigen:
The code below shows the issue.
I will do a pull request for a possible fix

```python
import adaptive
from adaptive.learner import SKOptLearner
def F_1D(x):
    return x**2
def F_2D(x):
    return x[0]**2 + x[1]**2

# Works
learner = SKOptLearner(F_1D, dimensions=[(-20., 20.)])
runner = adaptive.runner.simple(learner, goal=lambda l: l.npoints > 10)
# Doesn't work
learner = SKOptLearner(F_2D, dimensions=[(-20., 20.),(-10., 10.)])
runner = adaptive.runner.simple(learner, goal=lambda l: l.npoints > 10)
```

Error output:

```console
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-1fecee9ed826> in <module>
     35 try:
     36     learner = SKOptLearner(F_2D, dimensions=[(-20., 20.),(-10., 10.)])
---> 37     runner = adaptive.runner.simple(learner, goal=lambda l: l.npoints > 10)
        global runner = None
        global adaptive.runner.simple = <function simple at 0xb1565f8c0>
        global learner = <adaptive.learner.skopt_learner.SKOptLearner object at 0x1a197df990>
        global goal = undefined
        global l = undefined
        global l.npoints = undefined
     38 except:
     39     text_file = open("exeption.txt", "w")

/usr/local/anaconda3/envs/conda-dicarlo-lab/lib/python3.7/site-packages/adaptive/runner.py in simple(learner=<adaptive.learner.skopt_learner.SKOptLearner object>, goal=<function <lambda>>)
    681         for x in xs:
    682             y = learner.function(x)
--> 683             learner.tell(x, y)
        learner.tell = <bound method SKOptLearner.tell of <adaptive.learner.skopt_learner.SKOptLearner object at 0x1a197df990>>
        x = [-3.866859822710609, 2.529043603793383]
        y = 21.34866643838174
    684 
    685 

/usr/local/anaconda3/envs/conda-dicarlo-lab/lib/python3.7/site-packages/adaptive/learner/skopt_learner.py in tell(self=<adaptive.learner.skopt_learner.SKOptLearner object>, x=[-3.866859822710609, 2.529043603793383], y=21.34866643838174, fit=True)
     30 
     31     def tell(self, x, y, fit=True):
---> 32         self.pending_points.discard(x)
[...]

PR #234, opened 2019-12-05
Fix SKOptLearner for multi variate domain (issue #233)

caenrigen:
This fixes issue #233 

Tests checked. Per-commit checked.
Implemented additional test for this case.

Changed `data` type to `OrderedDict` to comply with other learners.

### DIFF
--- a/adaptive/learner/skopt_learner.py
+++ b/adaptive/learner/skopt_learner.py
@@ -31,7 +31,7 @@ class SKOptLearner(Optimizer, BaseLearner):
         super().__init__(**kwargs)
 
     def tell(self, x, y, fit=True):
-        if hasattr(x, '__iter__'):
+        if hasattr(x, "__iter__"):
             self.pending_points.discard(tuple(x))
             self.data[tuple(x)] = y
             super().tell(x, y, fit)
--- a/adaptive/learner/skopt_learner.py
+++ b/adaptive/learner/skopt_learner.py
@@ -26,24 +26,24 @@ class SKOptLearner(Optimizer, BaseLearner):
 
     def __init__(self, function, **kwargs):
         self.function = function
-        self.pending_points = list()
+        self.pending_points = set()
         self.data = OrderedDict()
         super().__init__(**kwargs)
 
     def tell(self, x, y, fit=True):
-        if x in self.pending_points:
-            self.pending_points.remove(x)
         if hasattr(x, '__iter__'):
+            self.pending_points.discard(tuple(x))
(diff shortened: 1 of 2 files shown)


caenrigen:
Strange that the `linting_test` failed
I get not error when running locally `pre-commit run --all-files`

How can it be fixed?

jbweston:
> Strange that the `linting_test` failed
> I get no error when running locally `pre-commit run --all-files`
> 
> How can it be fixed?

When I click through and look at the test failure I see that it's because `black` reformatted some files.
I am not sure why `pre-commit` succeeds on your machine. When I check out your branch locally and run `pre-commit run --all-files` the `black` check indeed fails.

caenrigen:
> > Strange that the `linting_test` failed
> > I get no error when running locally `pre-commit run --all-files`
> > How can it be fixed?
> 
> When I click through and look at the test failure I see that it's because `black` reformatted some files.
> 
> I am not sure why `pre-commit` succeeds on your machine. When I check out your branch locally and run `pre-commit run --all-files` the `black` check indeed fails.

You are right there was some message about black the first time I run pre-commit, I didn't notice it, I see the formatting changed in the file and it doesn't complain anymore.
Not sure how things work, if I do a new commit and push to the same branch you will be able to merge it?

adaptive/learner/skopt_learner.py
```diff
@@ -26,18 +28,23 @@ class SKOptLearner(Optimizer, BaseLearner):
     def __init__(self, function, **kwargs):
         self.function = function
         self.pending_points = set()
-        self.data = {}
+        self.data = collections.OrderedDict()
```
basnijholt: In Python >= 3.6 all dicts are ordered in principle, but it is good to make it explicit.

basnijholt:
@jbweston, should we just merge this into master? Seems like we didn't put anything on https://github.com/python-adaptive/adaptive/tree/stable-0.9.

caenrigen:
> LGTM! Thanks a lot for your contribution!

Thank you, guys, for the nice package! 😃

jbweston:
> @jbweston, should we just merge this into master? Seems like we didn't put anything on https://github.com/python-adaptive/adaptive/tree/stable-0.9.

Depends; do we want to release v0.9.1? So far we have not been properly adhering to semver, which I think is fine in our pre-1.0 no-backwards-compatibility state. I think we can elide keeping `stable-*` branches around always put things into the next minor release.

basnijholt:
I think we can skip a 0.9.1. So let's merge!

PR #236, opened 2019-12-11
add long_description to setup.py

basnijholt:
See title ^

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,8 @@ def get_version_and_cmdclass(package_name):
 
 version, cmdclass = get_version_and_cmdclass("adaptive")
 
+with open("README.rst") as f:
+    readme = f.read()
 
 install_requires = [
     "scipy",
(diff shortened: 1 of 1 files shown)


akhmerov:
Did you run `twine check dist/*`?

basnijholt:
Now I did, it uncovered a bug...

```
Checking dist/adaptive-0.10.0.dev14+g250ecbb-py3-none-any.whl: FAILED
  `long_description` has syntax errors in markup and would not be rendered on PyPI.
    line 64: Warning: "raw" directive disabled.
Checking dist/adaptive-0.10.0.dev14+g250ecbb.tar.gz: FAILED
  `long_description` has syntax errors in markup and would not be rendered on PyPI.
    line 64: Warning: "raw" directive disabled.
```

basnijholt:
It is because of the html in `Examples`.

PR #237, opened 2019-12-11
Power up pre-commit

basnijholt:
This adds some whitespace checks, `isort`, `pyupgrade`, and some other code checks from the `pre-commit-hooks` repo.

Note that `pyupgrade` doesn't really change much here because I used this already before.

A summary of what the hooks do:
* `pyupgrade` - A tool (and pre-commit hook) to automatically upgrade syntax for newer versions of the language.
* `end-of-file-fixer` - Makes sure files end in a newline and only a newline.
* `check-docstring-first` - Checks for a common error of placing code before the docstring.
* `check-yaml` - Attempts to load all yaml files to verify syntax.
* `debug-statements` - Check for debugger imports and py37+ breakpoint() calls in python source.
* `check-ast` - Simply check whether files parse as valid python.
* `seed-isort-config` - Statically populate the `known_third_party` `isort` setting because `isort` sucks at figuring out which third party libs are used. (from author of `pre-commit`)

I have also added this summary to the first commit.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,2 @@
+[settings]
+known_third_party = PIL,atomicwrites,holoviews,ipykernel,matplotlib,nbconvert,numpy,pytest,scipy,setuptools,skopt,sortedcollections,sortedcontainers,zmq
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,16 +1,43 @@
 repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.4.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-docstring-first
+    -   id: check-yaml
+    -   id: debug-statements
+    -   id: check-ast
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: 3.7.9
+    hooks:
+    -   id: flake8
+        args:
+          - --max-line-length=500
+          - --ignore=E203,E266,E501,W503
+          - --max-complexity=18
+          - --select=B,C,E,F,W,T4,B9
 -   repo: https://github.com/ambv/black
-    rev: 19.3b0
+    rev: 19.10b0
     hooks:
     - id: black
       language_version: python3.7
--   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.1.0
+-   repo: https://github.com/asottile/pyupgrade
+    rev: v1.25.2
     hooks:
-    -   id: end-of-file-fixer
-    -   id: trailing-whitespace
--   repo: https://gitlab.com/pycqa/flake8
-    rev: 3.7.8
+    -   id: pyupgrade
+        args: ['--py36-plus']
+-   repo: https://github.com/pre-commit/mirrors-isort
+    rev: v4.3.21
     hooks:
-    -   id: flake8
-        args: ['--max-line-length=500', '--ignore=E203,E266,E501,W503', '--max-complexity=18', '--select=B,C,E,F,W,T4,B9']
+    -   id: isort
+        args:
+          - --multi-line=3
+          - --trailing-comma
+          - --force-grid-wrap=0
+          - --use-parentheses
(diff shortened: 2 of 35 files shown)


.isort.cfg
```diff
@@ -0,0 +1,2 @@
+[settings]
+known_third_party = PIL,atomicwrites,holoviews,ipykernel,matplotlib,nbconvert,numpy,pytest,scipy,setuptools,skopt,sortedcollections,sortedcontainers,zmq
```
jbweston: what does this do?

Will this mean that if we ever add another library to the dependencies we'll need to add it here too?

.pre-commit-config.yaml
```diff
@@ -1,16 +1,43 @@
...
+-   repo: https://github.com/ambv/black
+    rev: 19.10b0
+    hooks:
+    - id: black
+      language_version: python3.7
+-   repo: https://github.com/asottile/pyupgrade
+    rev: v1.25.2
+    hooks:
+    -   id: pyupgrade
+        args: ['--py36-plus']
```
jbweston: Can you either doctor the commit message or add a comment about what each of these things does? e.g. I don't know what `pyupgrade` is about

.isort.cfg
basnijholt: `isort` isn't very good at figuring out what libraries are third party libs. The pre-commit hook `seed-isort-config` automatically maintains this file.

.pre-commit-config.yaml
basnijholt: See my updated comment [here](https://github.com/python-adaptive/adaptive/pull/237#issue-351901333)

basnijholt: and https://github.com/python-adaptive/adaptive/pull/237/commits/0a5cbe1d40a6b11a99c1b5262b411fc43d936db4

basnijholt:
isort is [able to deduce the requirements from `requirements.txt`](https://github.com/timothycrosley/isort/blob/c23f76f8ed020455a78d9480ef9e52f476cf601b/isort/finders.py#L306-L317). We could also define that and get rid of the `.isort.cfg` file.

.isort.cfg
basnijholt: `isort` is [able to deduce the requirements from `requirements.txt`](https://github.com/timothycrosley/isort/blob/c23f76f8ed020455a78d9480ef9e52f476cf601b/isort/finders.py#L306-L317). We could also define that and get rid of the `.isort.cfg` file.

jbweston: I was under the impression that `requirements.txt` was really for deploying apps, and not really for library development.

How does `seed-isort-config` work?

basnijholt: It [reads the source files](https://github.com/asottile/seed-isort-config/blob/master/seed_isort_config.py) IIUC.

jbweston: If `seed-isort-config` maintains this file, then we shouldn't check it in to git, right? It will just be autogenerated by `pre-commit`

basnijholt: It's not possible to not commit the file:
```bash
(base) basnijholt-macbook  ➜  adaptive git:(power-up-pre-commit) ✗  rm .isort.cfg
(base) basnijholt-macbook  ➜  adaptive git:(power-up-pre-commit) ✗  pre-commit run --all
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check docstring is first.................................................Passed
Check Yaml...............................................................Passed
Debug Statements (Python)................................................Passed
Check python ast.........................................................Passed
flake8...................................................................Passed
black....................................................................Passed
pyupgrade................................................................Passed
isort....................................................................Passed
seed isort known_third_party.............................................Failed
hookid: seed-isort-config

Files were modified by this hook. Additional output:

Creating an .isort.cfg with a known_third_party setting. Feel free to move the setting to a different config file in one of .editorconfig, .isort.cfg, setup.cfg, tox.ini, pyproject.toml.

This setting should be committed.
```

jbweston: never mind; I just tried it locally and it complains that `.isort.cfg` should be committed.

PR #247, opened 2019-12-16
use tox, closes #238

### LINKED ISSUE #238: use tox for testing
tox has a couple of benefits:

+ transparent testing against python/dependency version matrices
+ a single configuration file (many other tools in our testing pipeline e.g. flake8 read tox.ini)
+ can also build documentation, ensuring that the correct environment is used.

Check out https://github.com/jbweston/qsim/ to see how it looks.

basnijholt:
Adds tox.

## Summary of changes
* Use `tox.ini` for all configuration
* Test against Python 3.6, 3.7, and 3.8
* Tests on Windows, macOS, and Ubuntu
* Run tests with the bare dependencies (so only the bare requirements)
* Upload Coverage reports to [codecov.io](https://codecov.io/gh/python-adaptive/adaptive)
* Define dependencies all in `setup.py`
* Removes double occurrence of the pipelines
* Adds `flaky` to tests that sometimes fail
* Several small fixes in the tests
* Use `ProcessPoolExecutor` by default for Windows because it seems to work

Notes
* Reusing the Python env in different jobs [isn't possible](https://stackoverflow.com/questions/58322287/reuse-workspace-from-other-job-in-azure-pipelines) :( 
* ~~black fails because of https://github.com/psf/black/issues/1207~~ this is fixed

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,0 @@
-[flake8]
-max-line-length = 500
-ignore = E203,E266,E501,W503
-max-complexity = 18
-select = B,C,E,F,W,T4,B9
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,2 +1,0 @@
-[settings]
-known_third_party = PIL,atomicwrites,holoviews,ipykernel,matplotlib,nbconvert,numpy,pytest,scipy,setuptools,skopt,sortedcollections,sortedcontainers,zmq
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,26 +18,15 @@ repos:
     hooks:
     -   id: pyupgrade
         args: ['--py36-plus']
--   repo: https://github.com/pre-commit/mirrors-isort
-    rev: v4.3.21
-    hooks:
-    -   id: isort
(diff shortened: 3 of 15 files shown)


basnijholt:
@jbweston and @akhmerov, everything works now!

I have cleaned up the commits and added a summary in the first comment.

basnijholt:
We could use templates (adding an extra file), something like
```
parameters:
  pythonVersions: ['3.6', '3.7', '3.8'],
  operatingSystems: ['ubuntu-16.04', 'macOS-10.13', 'vs2017-win2016', 'windows-latest']

jobs:
- job: pytest
  strategy:
    matrix:
      ${{ each py in parameters.pythonVersions }}:
        ${{ each os in parameters.operatingSystems }}:
          ${{ format('{0}Python{1}', py, os) }}:
            python.version: '${{ py }}'
            vmImage: '${{ os }}'
```

However, this doesn't use Jinja2 and this would require also renaming the tox envs.

Let's not do it now and wait until https://github.com/microsoft/azure-pipelines-yaml/issues/20 is solved.

PR #248, opened 2019-12-16
added method to evaluate arbitrary domain points and tell the learner

caenrigen:
## Description
Hello, guys! I am using a learner as an optimiser (let me know if you are interested for me to do a pull request with that as well).

I felt the need to be able to evaluate specific points before executing the runner, in my case is a guess for the optimal domain region I am looking for.

Here is my implementation. Can it be added in this way?

I didn't implement any tests because I am not sure where it would be appropriate to include them.
## Checklist

- [X] Fixed style issues using `pre-commit run --all` (first install using `pip install pre-commit`)
- [X] `pytest` passed

## Type of change

- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/adaptive/learner/base_learner.py
+++ b/adaptive/learner/base_learner.py
@@ -1,4 +1,5 @@
 import abc
+from collections.abc import Iterable
 from contextlib import suppress
 from copy import deepcopy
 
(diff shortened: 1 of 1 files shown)


akhmerov:
Hmm, this seems to be mostly a shortcut saving a couple lines of code. We try to avoid adding those unless there is some nontrivial logic involved.

How do you use the learner as an optimizer? That sounds interesting.

caenrigen:
> Hmm, this seems to be mostly a shortcut saving a couple lines of code. We try to avoid adding those unless there is some nontrivial logic involved.
I understand, it should maybe be documented at least how to do it. I had to look in the source code to find the answer.
> How do you use the learner as an optimizer? That sounds interesting.

I limit the maximum resolution (minimum simplex volume) and use my cost function as the loss when the resolution limit is respected. More specifically, I average my cost function value over the simplex' edges, the averaging effect should provide some noise resilience which is needed in noise experiments that can confuse other optimisers. I keep normalising the values of my cost function to the (min, max) range seen so far. It keeps searching for better optima.

PR #254, opened 2020-01-05
add support for complex numbers in Learner1D

stuartthomas25:
## Description
I ran into an issue interpolating scattering matrices using adaptive where Learner1D.tell casts complex data to real. The reasoning for this is not immediately apparent so I made a small change that seems to work fine in this case. This is by no means a complete implementation but it seems to function fine with the loss function.

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/learner/learner1D.py
+++ b/adaptive/learner/learner1D.py
@@ -386,8 +386,11 @@ class Learner1D(BaseLearner):
             )
 
         # either it is a float/int, if not, try casting to a np.array
-        if not isinstance(y, (float, int)):
-            y = np.asarray(y, dtype=float)
+        if not isinstance(y, (float, int, complex)):
+            if np.iscomplexobj(y):
+                y = np.asarray(y, dtype=complex)
+            else:
+                y = np.asarray(y, dtype=float)
 
         # Add point to the real data dict
         self.data[x] = y


akhmerov:
Well spotted!

Hm, for sure the original code also looks like a bug (just silently discarding the complex part is never the  right thing to do).

basnijholt:
Good catch! We should add a test that catches this problem too.

basnijholt:
I just remembered that in the past we have always recommended implementing functions that return complex numbers as a vector and add the logic in the loss function. Still, we shouldn't discard the user's input.

basnijholt:
I have added a test, @akhmerov or @jbweston, feel free to merge.

akhmerov:
How do we treat the rescaling for complex data? I expect that we apply `np.ptp` somewhere, and that might not work correctly.

basnijholt:
Damn, I hadn't thought about that.

I am tempted to just raise an error when using complex numbers, and instead recommend people to use vectors.

jbweston:
@basnijholt status update / are we blocked?

akhmerov:
We could implement some rescaling policy that we consider to be reasonably well suited to complex numbers. For example we could choose the overall scale to be scalar, and rescale the values so that both the real and imaginary parts lie within `[0, 1]` interval.

basnijholt:
I think we should close this because this is better than done using a vector. We should document how to do this though. Probably the FAQ would be a good place to do this.

Otherwise, properly implementing this is probably much more complicated than what this PR does.

akhmerov:
@basnijholt perhaps, but the current rescaling logic doesn't suit the case of complex numbers. For example, if we are computing something that's real down to machine precision, the learner will amplify the noise. Choosing the same scale for real and imaginary parts is more appropriate.

basnijholt:
But someone can just do that using the right loss function.

Davide-sd:
> I just remembered that in the past we have always recommended implementing functions that return complex numbers as a vector and add the logic in the loss function.

> I think we should close this because this is better than done using a vector. We should document how to do this though. Probably the FAQ would be a good place to do this.

Hello Bas, I looked over the docs but I haven't found anything about dealing with complex numbers. Would you be able to provide an example?

PR #264, opened 2020-04-09
make learners picklable

basnijholt:
## Description

I realized that in some cases it's very useful to pickle learners. For example to send it over the network when parallelizing code.

With these changes, most learners become picklable.

## Checklist

- [ ] `pytest` passed

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/tests/test_pickling.py
+++ b/adaptive/tests/test_pickling.py
@@ -85,6 +85,8 @@ def test_serialization_for(learner_type, learner_kwargs, serializer):
         f = f_for_pickle  # noqa: F811
 
     learner = learner_type(f, **learner_kwargs)
+    if learner_type is Learner1D:
+        learner._recompute_losses_factor = 1
 
     simple(learner, goal_1)
     learner_bytes = serializer.dumps(learner)
--- a/adaptive/tests/test_pickling.py
+++ b/adaptive/tests/test_pickling.py
@@ -86,6 +86,8 @@ def test_serialization_for(learner_type, learner_kwargs, serializer):
 
     simple(learner, goal_1)
     learner_bytes = serializer.dumps(learner)
+    loss = learner.loss()
+    asked = learner.ask(1)
 
     if serializer is not pickle:
         # With pickle the functions are only pickled by reference
@@ -94,6 +96,15 @@ def test_serialization_for(learner_type, learner_kwargs, serializer):
 
     learner_loaded = serializer.loads(learner_bytes)
     assert learner_loaded.npoints >= 10
+    assert loss == learner_loaded.loss()
+
+    if learner_type is not Learner2D:
+        # cannot test this for Learner2D because
+        # xfailing test_point_adding_order_is_irrelevant
+        assert asked == learner_loaded.ask(1)
+        # load again to undo the ask
+        learner_loaded = serializer.loads(learner_bytes)
+
     simple(learner_loaded, goal_2)
     assert learner_loaded.npoints >= 20
 
--- a/adaptive/tests/test_pickling.py
+++ b/adaptive/tests/test_pickling.py
@@ -1,6 +1,4 @@
-import operator
 import pickle
-import random
 
 import pytest
 
@@ -11,7 +9,6 @@ from adaptive.learner import (
     IntegratorLearner,
     Learner1D,
     Learner2D,
-    LearnerND,
     SequenceLearner,
 )
 from adaptive.runner import simple
@@ -39,48 +36,61 @@ def goal_2(learner):
     return learner.npoints == 20
 
 
+def pickleable_f(x):
+    return hash(str(x)) / 2 ** 63
+
+
+nonpickleable_f = lambda x: hash(str(x)) / 2 ** 63  # noqa: E731
+
+
+def identity_function(x):
+    return x
+
+
+def datasaver(f, learner_type, learner_kwargs):
+    return DataSaver(
+        learner=learner_type(f, **learner_kwargs), arg_picker=identity_function
+    )
+
+
+def balancing_learner(f, learner_type, learner_kwargs):
+    learner_1 = learner_type(f, **learner_kwargs)
+    learner_2 = learner_type(f, **learner_kwargs)
+    return BalancingLearner([learner_1, learner_2])
+
+
 learners_pairs = [
     (Learner1D, dict(bounds=(-1, 1))),
     (Learner2D, dict(bounds=[(-1, 1), (-1, 1)])),
-    (LearnerND, dict(bounds=[(-1, 1), (-1, 1), (-1, 1)])),
     (SequenceLearner, dict(sequence=list(range(100)))),
     (IntegratorLearner, dict(bounds=(0, 1), tol=1e-3)),
     (AverageLearner, dict(atol=0.1)),
+    (datasaver, dict(learner_type=Learner1D, learner_kwargs=dict(bounds=(-1, 1)))),
+    (
+        balancing_learner,
+        dict(learner_type=Learner1D, learner_kwargs=dict(bounds=(-1, 1))),
+    ),
 ]
 
-serializers = [pickle]
+serializers = [(pickle, pickleable_f)]
 if with_cloudpickle:
-    serializers.append(cloudpickle)
+    serializers.append((cloudpickle, nonpickleable_f))
 if with_dill:
-    serializers.append(dill)
+    serializers.append((dill, nonpickleable_f))
+
 
 learners = [
-    (learner_type, learner_kwargs, serializer)
-    for serializer in serializers
+    (learner_type, learner_kwargs, serializer, f)
+    for serializer, f in serializers
     for learner_type, learner_kwargs in learners_pairs
 ]
 
 
-def f_for_pickle(x):
-    return 1
-
-
-def f_for_pickle_datasaver(x):
-    return dict(x=x, y=x)
-
-
 @pytest.mark.parametrize(
-    "learner_type, learner_kwargs, serializer", learners,
+    "learner_type, learner_kwargs, serializer, f", learners,
 )
-def test_serialization_for(learner_type, learner_kwargs, serializer):
+def test_serialization_for(learner_type, learner_kwargs, serializer, f):
     """Test serializing a learner using different serializers."""
-
-    def f(x):
-        return random.random()
-
-    if serializer is pickle:
-        # f from the local scope cannot be pickled
-        f = f_for_pickle  # noqa: F811
 
     learner = learner_type(f, **learner_kwargs)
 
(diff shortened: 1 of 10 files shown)


basnijholt:
@akhmerov and @jbweston, I wonder if we should also pickle `adaptive.__version__` and warn the user that if they are loading a learner that is pickled with another version, that there is no guarantee that stuff's correct?

basnijholt:
@akhmerov, good idea, I've added that in https://github.com/python-adaptive/adaptive/pull/264/commits/54017f38e01e509c256565abe32b1ddff03eed3d.

adaptive/tests/test_pickling.py
```diff
@@ -0,0 +1,169 @@
...
+    loss = learner.loss()
+    asked = learner.ask(1)
+
+    if serializer is not pickle:
+        # With pickle the functions are only pickled by reference
+        del f
+        del learner
+
+    learner_loaded = serializer.loads(learner_bytes)
+    assert learner_loaded.npoints >= 10
```
akhmerov, line 98: Why not an exact equality here?

basnijholt: Fixed!

akhmerov:
@basnijholt any idea why tests fail?

basnijholt:
The failure is intermittent and unrelated to these changes, also happens in master sometimes. 

It seems like the `distributed` test keeps getting stuck on Windows and MacOS.

akhmerov:
But it seems like serialization tests are failing?

image

basnijholt:
Oh, you are right! I assumed it were the same failures I had seen before.

I think that is fixed by https://github.com/python-adaptive/adaptive/pull/264/commits/bc652133b91673d5f6e14bde15810e5846ee6052.

adaptive/tests/test_pickling.py
```diff
@@ -0,0 +1,173 @@
...
+    def f(x):
+        return random.random()
+
+    if serializer is pickle:
+        # f from the local scope cannot be pickled
+        f = f_for_pickle  # noqa: F811
+
+    learner = learner_type(f, **learner_kwargs)
+    if learner_type is Learner1D:
+        learner._recompute_losses_factor = 1
```
akhmerov, line 89: If I understand correctly, here we're saying "we're only computing stuff approximately, so let's change that to make pickling work". I don't think that's the right way to go.

I think pickling is supposed to also store the internal state, so that we shouldn't need to tweak the learner behavior like that.

basnijholt: I don't think it is necessary to require that an unpickled learner is identical to the one before. In this case, the loaded learner has a better approximation of the loss because it is able to load all points at once.

Storing all tiny details will probably mean that with the tiniest of changes (to private API), one is not able to load saved learners from a previous version.

akhmerov: > In this case, the loaded learner has a better approximation of the loss because it is able to load all points at once.

That is a possible reason to deviate from the standard behavior, however, I'd also like to hear @jbweston's opinion—I still think that this is dangerously unconventional.

> Storing all tiny details will probably mean that with the tiniest of changes (even to private API), one is not able to load saved learners from a previous version.

I don't think cross-version pickling is guaranteed by anything.

basnijholt: Even today I loaded data saved with Adaptive more than three years ago and it loaded without any complaint, which made me very happy. 

One other thing to keep in mind is that, even though this might not be the final answer, it's still better than the current broken state! Right now picking works for some Learners while others have a broken state after loading.

akhmerov: That's the difference between data formats and pickling (preserving live objects).

One should be stable, the other should be exact.

jbweston: It is generally true that people expect `pickle.loads(pickle.dumps(x))` to be as close to the identity as possible. If we start deviating from this it is IMO almost certain that we (or one of our users) will get surprised by this at some point in the future.

I would prefer that if what someone wants to do is "create a learner with the same data" then they should export the data from the learner, pickle that, and then load it back in later.

jbweston: I am going to counter the "but this is functionally the same thing" response with "no". 

There are certain semantics associated with pickling/unpickling that people expect.

We do not set up the expectation that exporting/importing data from a learner has any given semantics beyond the most trivial (e.g. that after the import the learner has all the data specified).

basnijholt: OK fair enough, I've made the appropriate changes such that the data structures are the same!

I've run the test 10_000 times and it always passes now! 🎉

adaptive/tests/test_pickling.py
```diff
@@ -0,0 +1,169 @@
...
+def test_serialization_for_datasaver(serializer):
+    def f(x):
+        return dict(x=1, y=x ** 2)
+
+    if serializer is pickle:
+        # f from the local scope cannot be pickled
+        f = f_for_pickle_datasaver  # noqa: F811
+
+    _learner = Learner1D(f, bounds=(-1, 1))
+    learner = DataSaver(_learner, arg_picker=operator.itemgetter("y"))
```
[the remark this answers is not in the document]
basnijholt: Completely refactored this.

adaptive/tests/test_pickling.py
```diff
@@ -0,0 +1,169 @@
...
+
+    simple(learner_loaded, goal_2)
+    assert learner_loaded.npoints == 20
+
+
+@pytest.mark.parametrize(
+    "serializer", serializers,
+)
+def test_serialization_for_datasaver(serializer):
+    def f(x):
```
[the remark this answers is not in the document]
basnijholt: Everything is now specified in the `learners` list.

adaptive/tests/test_pickling.py
```diff
@@ -0,0 +1,169 @@
...
+    learner = learner_type(f, **learner_kwargs)
+
+    simple(learner, goal_1)
+    learner_bytes = serializer.dumps(learner)
+    loss = learner.loss()
+    asked = learner.ask(10)
+    data = learner.data
+
+    if serializer is not pickle:
+        # With pickle the functions are only pickled by reference
```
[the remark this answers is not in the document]
basnijholt: After refactor this special treatment of pickle isn't needed anymore.

adaptive/tests/test_pickling.py
```diff
@@ -0,0 +1,169 @@
...
+)
+def test_serialization_for(learner_type, learner_kwargs, serializer):
+    """Test serializing a learner using different serializers."""
+
+    def f(x):
+        return random.random()
+
+    if serializer is pickle:
+        # f from the local scope cannot be pickled
+        f = f_for_pickle  # noqa: F811
```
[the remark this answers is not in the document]
basnijholt: Done!

adaptive/tests/test_pickling.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+@pytest.mark.parametrize(
+    "learner_type, learner_kwargs, serializer", learners,
+)
+def test_serialization_for(learner_type, learner_kwargs, serializer):
+    """Test serializing a learner using different serializers."""
+
+    def f(x):
+        return random.random()
```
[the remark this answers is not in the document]
basnijholt: Fixed.

adaptive/tests/test_pickling.py
```diff
@@ -0,0 +1,169 @@
...
+
+learners = [
+    (learner_type, learner_kwargs, serializer)
+    for serializer in serializers
+    for learner_type, learner_kwargs in learners_pairs
+]
+
+
+def f_for_pickle(x):
+    return 1
```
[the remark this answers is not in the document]
basnijholt: Fixed.

adaptive/tests/test_pickling.py
```diff
@@ -0,0 +1,169 @@
...
+    learner_loaded = serializer.loads(learner_bytes)
+    assert learner_loaded.npoints == 10
+    simple(learner_loaded, goal_2)
+    assert learner_loaded.npoints == 20
+
+
+@pytest.mark.parametrize(
+    "serializer", serializers,
+)
+def test_serialization_for_balancing_learner(serializer):
```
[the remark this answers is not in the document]
basnijholt: Fixed.

adaptive/learner/learnerND.py
```diff
@@ -1176,3 +1169,23 @@ def _get_plane_color(simplex):
             opacity=opacity,
             lighting=lighting,
         )
+
+    def _get_data(self):
+        return self.data
+
+    def _set_data(self, data):
+        if data:
+            self.tell_many(*zip(*data.items()))
```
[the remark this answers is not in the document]
basnijholt: You are right!

I will do the work to make the LearnerND pickleable in https://github.com/python-adaptive/adaptive/pull/272.

adaptive/learner/learnerND.py
```diff
@@ -1176,3 +1169,23 @@ def _get_plane_color(simplex):
...
+            self.function,
+            self.bounds,
+            self.loss_per_simplex,
+            self._get_data(),
+        )
+
+    def __setstate__(self, state):
+        function, bounds, loss_per_simplex, data = state
+        self.__init__(function, bounds, loss_per_simplex)
+        self._set_data(data)
```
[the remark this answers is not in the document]
basnijholt: I will do the work to make the LearnerND pickleable in https://github.com/python-adaptive/adaptive/pull/272.

tlaeven:
Using a cached function doesn't work in this branch, is this intentional?

#### Minimal example:
```python
import adaptive
from functools import lru_cache
adaptive.notebook_extension()

@lru_cache()
def g(x):
    return x**2

def f(x):
    return x-g(0)

learner = adaptive.Learner1D(f, [-1,1])

runner = adaptive.Runner(learner)
runner.live_info()
```

Raises an error:
`loky.process_executor.BrokenProcessPool: A task has failed to un-serialize. Please ensure that the arguments of the function are all picklable.`

basnijholt:
@tlaeven, this is actually unrelated to these changes. The code you posted never worked, and has nothing to do with Adaptive:
```python
from functools import lru_cache
import loky
from concurrent.futures import ProcessPoolExecutor

@lru_cache()
def g(x):
    return x ** 2

def f(x):
    return x - g(0)

# Before
with ProcessPoolExecutor() as ex:
    fut = ex.submit(f, 0)
    try:
        fut.result()
    except Exception as e:
        print(f"ProcessPoolExecutor failed: {e}")

# Now
ex = loky.get_reusable_executor()
fut = ex.submit(f, 0)
try:
    fut.result()
except Exception as e:
    print(f"loky failed: {e}")
```

basnijholt:
@jbweston, I've fixed all that was brought up.

The `LearnerND` takes some more work, so I moved those commits to https://github.com/python-adaptive/adaptive/pull/272.

I think it's ready to merge. Would you take a final look?

Issue #265, opened 2020-04-14
SequenceLearner with unhashable entries in sequence break Runner


basnijholt:
When a `SequenceLearner` that has unhashable entries in sequence (e.g., `List[dict]`) fails, the `Runner` breaks.

This happens because of https://github.com/python-adaptive/adaptive/blob/0f85d3abd1cb0eddf195dec0deabd6d0dd97d268/adaptive/runner.py#L204-L210

Reproduce with:

```python
import adaptive
import numpy as np

adaptive.notebook_extension()

def f(dct):
    import random

    if random.random() < 0.5:
        raise Exception()
    return dct["x"]


seq = [dict(x=x) for x in np.linspace(-1, 1, 101)]  # unhashable

learner = adaptive.SequenceLearner(f, sequence=seq)

runner = adaptive.Runner(learner, goal=adaptive.SequenceLearner.done)
runner.live_info()
```

```python
runner.task.print_stack()
```

traceback
```
Traceback for <Task finished name='Task-3' coro=<AsyncRunner._run() done, defined at /Users/basnijholt/Work/adaptive/adaptive/runner.py:601> exception=TypeError("unhashable type: 'dict'")> (most recent call last):
TypeError: unhashable type: 'dict'
  File "/Users/basnijholt/Work/adaptive/adaptive/runner.py", line 555, in status
    return "failed"
  File "/Users/basnijholt/Work/adaptive/adaptive/runner.py", line 555, in status
    return "failed"
  File "/Users/basnijholt/Work/adaptive/adaptive/runner.py", line 618, in _run
    self._cleanup()
  File "/Users/basnijholt/Work/adaptive/adaptive/runner.py", line 210, in _process_futures
    self._do_raise(e, x)

```

PR #266, opened 2020-04-14
Make SequenceLearner points hashable by passing the sequence to the function.

basnijholt:
## Description

With this change, the function of the `SequenceLearner` contains the sequence.

This is needed because points that are passed to the `Runner` **need** to be hashable, and this is not required by the entries of the `sequence` in `SequenceLearner`.

Fixes https://github.com/python-adaptive/adaptive/issues/265

## Checklist

- [x] `pytest` passed

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/learner/sequence_learner.py
+++ b/adaptive/learner/sequence_learner.py
@@ -6,14 +6,7 @@ from adaptive.learner.base_learner import BaseLearner
 
 
 class _IndexToPoint:
-    """Call function with index of sequence.
-
-    The SequenceLearner's function receives a tuple ``(index, point)``
-    but the original function only takes ``point``.
-
-    This is the same as `lambda x: function(x[1])`, however, that is not
-    pickable.
-    """
+    """Call function with index of sequence."""
 
     def __init__(self, function, sequence):
         self.function = function
--- a/adaptive/learner/sequence_learner.py
+++ b/adaptive/learner/sequence_learner.py
@@ -5,7 +5,7 @@ from sortedcontainers import SortedDict, SortedSet
 from adaptive.learner.base_learner import BaseLearner
 
 
-class _IndexToPoint:
+class _CallFromSequence:
     """Call function with index of sequence."""
 
     def __init__(self, function, sequence):
@@ -45,7 +45,7 @@ class SequenceLearner(BaseLearner):
 
     def __init__(self, function, sequence):
         self._original_function = function
-        self.function = _IndexToPoint(function, sequence)
+        self.function = _CallFromSequence(function, sequence)
         self._to_do_indices = SortedSet({i for i, _ in enumerate(sequence)})
         self._ntotal = len(sequence)
         self.sequence = copy(sequence)
--- a/adaptive/learner/sequence_learner.py
+++ b/adaptive/learner/sequence_learner.py
@@ -5,8 +5,8 @@ from sortedcontainers import SortedDict, SortedSet
 from adaptive.learner.base_learner import BaseLearner
 
 
-class _IgnoreFirstArgument:
-    """Remove the first argument from the call signature.
+class _IndexToPoint:
+    """Call function with index of sequence.
 
     The SequenceLearner's function receives a tuple ``(index, point)``
     but the original function only takes ``point``.
@@ -15,18 +15,19 @@ class _IgnoreFirstArgument:
     pickable.
     """
 
-    def __init__(self, function):
+    def __init__(self, function, sequence):
         self.function = function
+        self.sequence = sequence
 
-    def __call__(self, index_point, *args, **kwargs):
-        index, point = index_point
+    def __call__(self, index, *args, **kwargs):
+        point = self.sequence[index]
         return self.function(point, *args, **kwargs)
 
     def __getstate__(self):
-        return self.function
+        return self.function, self.sequence
 
-    def __setstate__(self, function):
-        self.__init__(function)
+    def __setstate__(self, state):
+        self.__init__(*state)
 
 
 class SequenceLearner(BaseLearner):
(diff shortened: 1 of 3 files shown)


adaptive/learner/sequence_learner.py
```diff
@@ -15,18 +15,19 @@ class _IgnoreFirstArgument:
...
-    def __init__(self, function):
+    def __init__(self, function, sequence):
         self.function = function
+        self.sequence = sequence
 
-    def __call__(self, index_point, *args, **kwargs):
-        index, point = index_point
+    def __call__(self, index, *args, **kwargs):
+        point = self.sequence[index]
         return self.function(point, *args, **kwargs)
```
akhmerov, line 24: Maybe make a one-liner
```suggestion
        return self.function(self.sequence[point], *args, **kwargs)
```

adaptive/tests/test_sequence_learner.py
```diff
@@ -0,0 +1,24 @@
+import asyncio
+
+import numpy as np
+
+import adaptive
+
+
+def might_fail(dct):
```
akhmerov, line 8: This is more reliable and time-efficient.

```python
class FailOnce:
    def __init__(self):
        failed = False

    def __call__(self, value):
        if failed:
            return value
        failed = True
        raise RuntimeError
```

akhmerov:
I propose to:

- Repurpose #265 to an issue that runner breaks on unhashable inputs: that's the real problem.
- Make this PR dedicated to a simplification of the sequence learner (which happens to close the current #265 a side effect).
- Also throw away the unnecessary `__getstate__`/`__setstate__` logic from `_IndexToPoint`.

@basnijholt what do you think?

adaptive/learner/sequence_learner.py
```diff
@@ -5,8 +5,8 @@
 from adaptive.learner.base_learner import BaseLearner
 
 
-class _IgnoreFirstArgument:
-    """Remove the first argument from the call signature.
+class _IndexToPoint:
```
akhmerov, line 8: This name is a bit mysterious (sounds like a dict) as well as the docstring. Not sure what the better option could be.

adaptive/learner/sequence_learner.py
```diff
@@ -5,8 +5,8 @@
...
+    """Call function with index of sequence.
 
     The SequenceLearner's function receives a tuple ``(index, point)``
```
akhmerov, line 11: Obsolete docstring.

basnijholt:
@akhmerov, I've implemented the changes you suggested and opened https://github.com/python-adaptive/adaptive/issues/267.

akhmerov:
Sorry, I'm having second thoughts. Passing the complete sequence of arguments to the function means we're distributing it across all workers. Is that definitely OK? It might be, but I think we should consider this carefully.

adaptive/tests/test_sequence_learner.py
```diff
@@ -0,0 +1,28 @@
...
+
+class FailOnce:
+    def __init__(self):
+        self.failed = False
+
+    def __call__(self, value):
+        if self.failed:
+            return value
+        self.failed = True
+        raise Exception
```
akhmerov, line 17:
```suggestion
        raise RuntimeError
```
is more appropriate even in a mock.

adaptive/tests/test_sequence_learner.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def test_fail_with_sequence_of_unhashable():
+    # https://github.com/python-adaptive/adaptive/issues/265
+    seq = [dict(x=x) for x in np.linspace(-1, 1, 101)]  # unhashable
+    learner = SequenceLearner(FailOnce(), sequence=seq)
+    runner = Runner(
+        learner, goal=SequenceLearner.done, retries=100, executor=SequentialExecutor()
```
akhmerov, line 25:
```suggestion
        learner, goal=SequenceLearner.done, retries=1, executor=SequentialExecutor()
```

adaptive/tests/test_sequence_learner.py
```diff
@@ -0,0 +1,28 @@
...
+    def __call__(self, value):
+        if self.failed:
+            return value
+        self.failed = True
+        raise Exception
+
+
+def test_fail_with_sequence_of_unhashable():
+    # https://github.com/python-adaptive/adaptive/issues/265
+    seq = [dict(x=x) for x in np.linspace(-1, 1, 101)]  # unhashable
```
akhmerov, line 22:
```suggestion
    seq = [{1: 1}]  # unhashable
```

akhmerov:
> Passing the complete sequence of arguments to the function means we're distributing it across all workers. Is that definitely OK?

Ultimately I think this would only be a problem in extremely exotic cases, and the simplification of the codebase justifies this downside.

basnijholt:
I actually lean towards not accepting the overhead.

From learning the following function on a `ProcessPoolExecutor`:
```
class FailOnce:
    def __init__(self):
        self.n = 0

    def __call__(self, value):
        self.n += 1
        return self.n
```
Which will always return 1, I learn that the sequence will be serialized and send *each* point!

As an alternative solution, I have opened https://github.com/python-adaptive/adaptive/pull/268/files. Which appeared to be way less work than I thought initially.

basnijholt:
Superseded by https://github.com/python-adaptive/adaptive/pull/268.

akhmerov:
I'm actually not sure whether creating multiple copies of a sequence is an important limitation and I am wondering whether we should merge this anyway. I still believe this implementation of a sequence learner is simpler, and therefore deserves consideration separate of its relation to #265.

@jbweston, what do you think?

basnijholt:
@akhmerov, I've benchmarked this and the overhead is completely unacceptable.

For example, at this exact moment I am doing simulations with `SequenceLearners` of ~1e6 items with 15 parameters.

```
import pickle
import numpy as np
sequence = list(np.random.rand(1_000_000, 15))
%timeit pickle.dumps(sequence)
```
`4.92 s ± 44.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`
An that happens on **each** function evaluation. And this does not even include unserializing and sending it over the network.

akhmerov:
OK, fair enough.

PR #268, opened 2020-04-16
make the Runner work with unhashable points

### LINKED ISSUE #267: Make Runner work with unhashable points
Currently `Runner`s do not support the points to be unhashable. This leads to https://github.com/python-adaptive/adaptive/issues/265.

Allowing the `Runner` to work with any points would simplify the `SequenceLearner` and means we wouldn't have to convert numpy arrays to tuples in all other learners.

basnijholt:
## Description

Enable the `Runner` to with with unhashable points.

Fixes #267.

## Checklist

- [ ] `pytest` passed

## Type of change

*Check relevant option(s).*

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -55,9 +55,7 @@ _default_executor = (
 
 
 def _key_by_value(dct, value):
-    for k, v in dct.items():
-        if v == value:
-            return k
+    return next((k for k, v in dct.items() if v == value), None)
 
 
 class BaseRunner(metaclass=abc.ABCMeta):
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -56,10 +56,6 @@ _default_executor = (
 )
 
 
-def _key_by_value(dct, value):
-    return next(k for k, v in dct.items() if v == value)
-
-
 class BaseRunner(metaclass=abc.ABCMeta):
     r"""Base class for runners that use `concurrent.futures.Executors`.
 
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -174,22 +174,21 @@ class BaseRunner(metaclass=abc.ABCMeta):
         return self.log is not None
 
     def _ask(self, n):
-        points = []
-        for i, pid in enumerate(self._to_retry.keys()):
-            if i == n:
-                break
-            point = self._id_to_point[pid]
-            if point not in self.pending_points.values():
-                points.append(point)
+        pids = [
+            pid
+            for pid in self._to_retry.keys()
+            if pid not in self.pending_points.values()
+        ][:n]
+        loss_improvements = len(pids) * [float("inf")]
 
-        loss_improvements = len(points) * [float("inf")]
-        if len(points) < n:
-            new_points, new_losses = self.learner.ask(n - len(points))
-            points += new_points
+        if len(pids) < n:
+            new_points, new_losses = self.learner.ask(n - len(pids))
             loss_improvements += new_losses
-            for p in new_points:
-                self._id_to_point[self._next_id()] = p
-        return points, loss_improvements
+            for point in new_points:
+                pid = self._next_id()
+                self._id_to_point[pid] = point
+                pids.append(pid)
+        return pids, loss_improvements
 
     def overhead(self):
         """Overhead of using Adaptive and the executor in percent.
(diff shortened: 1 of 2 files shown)


adaptive/runner.py
```diff
@@ -54,6 +54,10 @@
 )
 
 
+def _key_by_value(dct, value):
+    return next((k for k, v in dct.items() if v == value), None)
```
basnijholt, line 58: @akhmerov, do you think it would be better to raise `StopIteration` because now it's not possible to use `None` as a key.

adaptive/runner.py
```diff
@@ -198,20 +212,22 @@ def overhead(self):
     def _process_futures(self, done_futs):
         for fut in done_futs:
             x = self.pending_points.pop(fut)
+            i = _key_by_value(self._index_to_point, x)  # O(N)
```
basnijholt, line 215: In general, we can expect that the very first value in `self._index_to_point` will be `x` because it is simply the oldest/longest running point.

So I really don't think this operation will hurt.

@jbweston, do you have an opinion here?

jbweston:
I've scheduled time later this morning (pacific time) to take a look at this.

adaptive/runner.py
jbweston: If we change `pending_points` to store point IDs instead then we don't need this reverse lookup at all.

adaptive/runner.py
```diff
@@ -232,6 +248,12 @@ def _get_futures(self):
             fut = self._submit(x)
             fut.start_time = start_time
             self.pending_points[fut] = x
+            try:
+                i = _key_by_value(self._index_to_point, x)  # O(N)
+            except StopIteration:  # `x` is not a value in `self._index_to_point`
+                self._i += 1
+                i = self._i
+            self._index_to_point[i] = x
```
jbweston, line 256: This logic should be placed in `self._ask` instead, and can be simplified by having an ID-generating state machine as I show above

adaptive/runner.py
```diff
@@ -162,9 +171,14 @@ def do_log(self):
...
-            p for p in self.to_retry.keys() if p not in self.pending_points.values()
-        ][:n]
+        points = []
+        for i, index in enumerate(self._to_retry.keys()):
+            if i == n:
+                break
+            point = self._index_to_point[index]
+            if point not in self.pending_points.values():
+                points.append(point)
+
```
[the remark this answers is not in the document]
jbweston: Probably introduce `_pending_points` that uses point IDs and have a property `pending_points` that does the lookup, the same as we have for `tracebacks` etc.

adaptive/runner.py
```diff
@@ -143,14 +147,19 @@ def __init__(
...
         self.retries = retries
         self.raise_if_retries_exceeded = raise_if_retries_exceeded
-        self.to_retry = {}
-        self.tracebacks = {}
+        self._to_retry = {}
+        self._tracebacks = {}
+
+        # Keeping track of index -> point
+        self._index_to_point = {}
+        self._i = 0  # some unique index to be associated with each point
```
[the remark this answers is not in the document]
basnijholt: Good suggestion! Fixed in 2f71937e74b4038a908558d0b11fd3f419723438.

adaptive/runner.py
```diff
@@ -143,14 +147,19 @@ def __init__(
         # Error handling attributes
         self.retries = retries
         self.raise_if_retries_exceeded = raise_if_retries_exceeded
-        self.to_retry = {}
-        self.tracebacks = {}
+        self._to_retry = {}
+        self._tracebacks = {}
+
+        # Keeping track of index -> point
+        self._index_to_point = {}
```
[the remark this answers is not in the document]
basnijholt: Fixed in 20be4762f75070ec3193a3ff3f8a9d11dda0aab3.

adaptive/runner.py
basnijholt: Fixed in 62e9a416a964f43ab79b78c1463f67c04e28633e.

adaptive/runner.py
```diff
@@ -162,9 +171,14 @@ def do_log(self):
...
+        for i, index in enumerate(self._to_retry.keys()):
+            if i == n:
+                break
+            point = self._index_to_point[index]
+            if point not in self.pending_points.values():
+                points.append(point)
+
         loss_improvements = len(points) * [float("inf")]
         if len(points) < n:
             new_points, new_losses = self.learner.ask(n - len(points))
```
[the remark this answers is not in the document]
basnijholt: Done in 8f6edfd.

adaptive/runner.py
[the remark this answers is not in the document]
basnijholt: Fixed in 8f6edfd and 92f9902.

basnijholt:
@jbweston, thanks a lot for your review!

I've implemented all suggestions you've made :tada:

adaptive/runner.py
```diff
@@ -128,7 +128,7 @@ def __init__(
 
         self._max_tasks = ntasks
 
-        self._pending_points = {}
+        self._pending_pids = {}
```
akhmerov, line 131: `pid` is commonly used for a "process id". I propose to adopt the terminology more similar to other parallel executors and call these "_pending_tasks".

basnijholt: It's point id here.

akhmerov: Indeed, I understand what it means. I should have explained myself clearer: I believe a lot of people when seeing "pid" will think it's a process id, and therefore I propose to rename this attribute.

basnijholt: I don't think it will be confusing but nonetheless I've made the suggested change 👍

adaptive/runner.py
```diff
@@ -162,15 +170,20 @@ def do_log(self):
...
-        points = [
-            p for p in self.to_retry.keys() if p not in self.pending_points.values()
-        ][:n]
-        loss_improvements = len(points) * [float("inf")]
-        if len(points) < n:
-            new_points, new_losses = self.learner.ask(n - len(points))
-            points += new_points
+        pending_ids = self._pending_tasks.values()
+        pids_gen = (pid for pid in self._to_retry.keys() if pid not in pending_ids)
+        pids = list(itertools.islice(pids_gen, n))
```
jbweston, line 175: If it is important to keep it like this, please comment about why a generator comprehension is used.

The common idiom would be to use a list comprehension

jbweston: if it is for readability only then a comment is not necessary, and you can just resolve with convo

basnijholt: I moved from the list comprehension because not the whole list needs to be created, only until `n`.

basnijholt: I added a comment!

Issue #273, opened 2020-05-14
add minimum number of points parameter to AverageLearner


basnijholt:
The `AverageLearner` samples a random 

Like in https://github.com/python-adaptive/adaptive/pull/143 we should have an argument
```
    min_seeds_per_point : int, default 3
        Minimum amount of values per point. This means that the
        standard error of a point is infinity until there are
        'min_seeds_per_point' for a point.
```

Right now, it can happen that 2 consecutive sampling events happen to be the same up to some very small precision, resulting in a converged learner.

For example, in my case I sampled 2000 `AverageLearner`s until convergence, most of them sampled ~3000 times except for this one:

![image](https://user-images.githubusercontent.com/6897215/81922029-cc870a80-95db-11ea-94e6-5217357923b1.png)

The problem can be demonstrated using the following code.

Find a seed that produces two numbers that are close
```python
for i in range(10_000_000):
    random.seed(i)
    r1 = random.random()
    r2 = random.random()
    if abs(r1 - r2) < 1e-7:
        break

random.seed(i)
print(i)
```
`3317884`

Then
```python
random.seed(3317884)

def f(_):
    return random.random()

learner = adaptive.AverageLearner(f, rtol=0.01, atol=0.01)
adaptive.runner.simple(learner, goal=lambda l: l.loss() < 1)
print(learner.npoints, learner.data)
```
`2 {0: 0.6232189044314053, 1: 0.6232188298033593}`

PR #274, opened 2020-05-14
AverageLearner: implement min_npoints

basnijholt:
## Description

Adds `min_npoints` as a parameter to the `AverageLearner`.

Closes https://github.com/python-adaptive/adaptive/issues/273

## Checklist

- [x] `pytest` passed

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/learner/average_learner.py
+++ b/adaptive/learner/average_learner.py
@@ -19,6 +19,8 @@ class AverageLearner(BaseLearner):
         Desired absolute tolerance.
     rtol : float
         Desired relative tolerance.
+    min_npoints : int
+        Minimum number of points required to estimate the standard deviation.
 
     Attributes
     ----------
@@ -30,7 +32,9 @@ class AverageLearner(BaseLearner):
         Number of evaluated points.
     """
 
-    def __init__(self, function, atol=None, rtol=None):
+    def __init__(self, function, atol=None, rtol=None, min_npoints=2):
+        if min_npoints < 2:
+            raise ValueError("`min_npoints` should be at least 2.")
         if atol is None and rtol is None:
             raise Exception("At least one of `atol` and `rtol` should be set.")
         if atol is None:
@@ -44,6 +48,7 @@ class AverageLearner(BaseLearner):
         self.atol = atol
         self.rtol = rtol
         self.npoints = 0
+        self.min_npoints = min_npoints
         self.sum_f = 0
         self.sum_f_sq = 0
 
@@ -92,7 +97,7 @@ class AverageLearner(BaseLearner):
         """The corrected sample standard deviation of the values
         in `data`."""
         n = self.npoints
-        if n < 2:
+        if n < self.min_npoints:
             return np.inf
         numerator = self.sum_f_sq - n * self.mean ** 2
         if numerator < 0:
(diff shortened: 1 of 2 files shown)


adaptive/tests/test_average_learner.py
```diff
@@ -46,3 +47,20 @@ def test_avg_std_and_npoints():
             assert learner.npoints == len(learner.data)
             assert abs(learner.sum_f - values.sum()) < 1e-13
             assert abs(learner.std - std) < 1e-13
+
+
+def test_min_npoints():
```
akhmerov, line 52: The test seems unnecessarily complex. I propose to use a fully constant function and check the number of points when done.

Or perhaps I don't understand why `npoints_similar` is relevant for the testing behavior.

adaptive/learner/average_learner.py
```diff
@@ -30,7 +32,9 @@ class AverageLearner(BaseLearner):
         Number of evaluated points.
     """
 
-    def __init__(self, function, atol=None, rtol=None):
+    def __init__(self, function, atol=None, rtol=None, min_npoints=2):
+        if min_npoints < 2:
+            raise ValueError("`min_npoints` should be at least 2.")
```
akhmerov, line 37: Should we really raise or just sneakily

```suggestion
        # Cannot estimate standard deviation with less than 2 points.
        min_npoints = max(min_npoints, 2)
```

instead?

adaptive/tests/test_average_learner.py
basnijholt: I simplified the test in https://github.com/python-adaptive/adaptive/pull/274/commits/02c4047fc705fb3b79a16e5b9cb2fd98897babd1. Good?

adaptive/learner/average_learner.py
basnijholt: I thought about that too but ultimately, I don't really have a preference.

Do you?

akhmerov: I think what I proposed is nicer: it's not that we don't respect the user settings if the user asks for at least fewer points than 2. And we don't have to document it if there's no error.

adaptive/learner/average_learner.py
```diff
@@ -19,6 +19,8 @@ class AverageLearner(BaseLearner):
         Desired absolute tolerance.
     rtol : float
         Desired relative tolerance.
+    min_npoints : int
+        Minimum number of points required to estimate the standard deviation.
```
akhmerov, line 23:
```suggestion
        Minimum number of points to sample.
```

adaptive/learner/average_learner.py
```diff
@@ -44,6 +46,8 @@ def __init__(self, function, atol=None, rtol=None):
         self.atol = atol
         self.rtol = rtol
         self.npoints = 0
+        # Cannot estimate standard deviation with less than 2 points.
```
akhmerov, line 49:
```suggestion
        # Cannot estimate standard deviation with fewer than 2 points.
```

Sorry, I'll be pedantic here.

adaptive/tests/test_average_learner.py
basnijholt: It could be even simpler with a constant function (like you said). Although I am not sure whether a reader of this test would understand the underlying problem that it's testing.

akhmerov: Let's simplify this further.

- We don't need actual randomness.
- We need to ensure that a function with which the learner would stop before reaching min_points runs for at least min_points. Therefore we need to assert that if we run without `min_points`, then 2 points are evaluated, and otherwise at least `min_points.

akhmerov:
Thanks, LGTM now.

PR #281, opened 2020-05-28
always serialize the function using cloudpickle

basnijholt:
## Description

This was a suggestion of @jbweston some time ago.

Using this, the learners are picklable regardless of the serialization package chosen.

This also means that we can e.g. use lambda functions when using a `ProcessPoolExecutor` and do not need `loky` per se.

I am making these changes now because I am finding that in some cases it's not preferable to use `cloudpickle` to serialize simple dictionaries (see https://github.com/cloudpipe/cloudpickle/issues/375).

## Checklist

- [ ] `pytest` passed

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/learner/average_learner.py
+++ b/adaptive/learner/average_learner.py
@@ -1,5 +1,6 @@
 from math import sqrt
 
+import cloudpickle
 import numpy as np
 
 from adaptive.learner.base_learner import BaseLearner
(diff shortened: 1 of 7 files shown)


ogrisel:
As noted in  cloudpipe/cloudpickle#375, the performance problem when serializing large collections (list, dicts, sets) of small object is caused by the fact that prior to Python 3.8, cloudpickle had not other choice than subclass the Python implementation of the `Pickler` class which is slow.

Since Python 3.8, cloudpickle can subclass the C implementation of the `Pickler` class which should fix this performance problem.

basnijholt:
@ogrisel, thanks a lot for your explanation (also in https://github.com/cloudpipe/cloudpickle/issues/375#issuecomment-635886978)!

PR #283, opened 2020-06-19
AverageLearner1D added

AlvaroGI:
The AverageLearner1D has been added to the master branch (original code with design details and tests can be found [here](https://github.com/AlvaroGI/adaptive/tree/AverageLearner2D/)).
Tutorial notebook added (tutorial_averagelearner1d_aux.py contains some auxiliary functions for the tutorial).

## Changes in existing files
The __init__.py files of adaptive/ and learner/, and notebook_integration.py were only modified to include the AverageLearner1D.


adaptive/learner/averagelearner1D.py
```diff
@@ -0,0 +1,399 @@
...
+            self._error_in_mean[x] = self._calc_error_in_mean(ys, y_avg, n)
+            self._update_distances(x)
+            self._update_rescaled_error_in_mean(x, "resampled")
+
+            if self._error_in_mean[x] <= self.min_Delta_g or n >= self.max_samples:
+                self._rescaled_error_in_mean.pop(x, None)
+
+            # We also need to update scale and losses
+            super()._update_scale(x, y)
+            self._update_losses_resampling(x, real=True)  #  XXX: REVIEW
```
basnijholt, line 266: @AlvaroGI, what does `REVIEW` mean here?

adaptive/learner/averagelearner1D.py
```diff
@@ -0,0 +1,399 @@
...
+
+            # XXX: Is the following necessary?
```
basnijholt, line 268: @AlvaroGI, do you know how to check whether this is needed?

adaptive/learner/averagelearner1D.py
```diff
@@ -0,0 +1,399 @@
...
+        if (b is not None) and right_loss_is_unknown:
+            self.losses_combined[x, b] = float("inf")
+
+    def _calc_error_in_mean(self, ys, y_avg, n):
+        variance_in_mean = sum((y - y_avg) ** 2 for y in ys) / (n - 1)
+        t_student = scipy.stats.t.ppf(1 - self.alpha, df=n - 1)
+        return t_student * (variance_in_mean / n) ** 0.5
+
+    def tell_many(self, xs, ys, *, force=False):
+        """The data should be given as:
```
basnijholt, line 319: doc-string is incorrect

adaptive/learner/averagelearner1D.py
```diff
@@ -0,0 +1,399 @@
...
+                - {x_i: y_i} (only the mean at each point), in which case the
+                  number of samples is assumed to be 1 and the error 0 for all
+                  data points. These points will not be included in
+                  _rescaled_error_in_mean and therefore will not be resampled.
+                - {x_i: [y_i0, y_i1, ...]} (all data samples at each point)."""
+        # If data_samples is given:
+        if isinstance(ys[0], list):
```
basnijholt, line 326: maybe should check for any iterable

adaptive/learner/averagelearner1D.py
```diff
@@ -0,0 +1,399 @@
...
+        hv = ensure_holoviews()
+        if not self.data:
+            p = hv.Scatter([]) * hv.ErrorBars([]) * hv.Path([])
+        elif not self.vdim > 1:
+            p = (
+                hv.Scatter(self.data)
+                * hv.ErrorBars(
+                    [(x, self.data[x], self._error_in_mean[x]) for x in self.data]
+                )
+                * hv.Path([])
```
basnijholt, line 381: adding a line through the points would probably be nicer.

adaptive/learner/averagelearner1D.py
AlvaroGI: Nothing anymore, it was a comment for myself and I forgot to remove it.

adaptive/learner/averagelearner1D.py
basnijholt: Fixed

adaptive/learner/averagelearner1D.py
```diff
@@ -0,0 +1,399 @@
...
+class AverageLearner1D(Learner1D):
+    """Learns and predicts a noisy function 'f:ℝ → ℝ^N'.
+
+    New parameters (wrt Learner1D)
+    ------------------------------
+    delta : float
+        This parameter controls the resampling condition. A point is resampled
+        if its uncertainty is larger than delta times the smallest neighboring
+        interval.
+        We strongly recommend 0 < delta <= 1.
```
basnijholt: Isn't this even required?

adaptive/learner/averagelearner1D.py
```diff
@@ -0,0 +1,399 @@
...
+            self._error_in_mean.update(zip(xs, [0] * len(xs)))
+            for x in xs:
+                self._update_distances(x)
+
+    def plot(self):
+        """Returns a plot of the evaluated data with error bars.
+
+        Returns
+        -------
+        plot : `holoviews.element.Scatter` (if vdim=1)\
```
basnijholt: the return type is incorrect

basnijholt:
Awesome!
AverageLearner1D mov

adaptive/learner/averagelearner1D.py
AlvaroGI: It is, I just removed this comment.

adaptive/learner/averagelearner1D.py
AlvaroGI: Not really. If delta<0, the algorithm will only resample existing points and stop sampling new ones. If delta>1, the algorithm will mostly sample new points (larger delta means less resampling). These two regimes are not commonly desired, but one may use them in some very specific cases. That is why I decided to allow for any value of delta, although in general we recommend 0<delta<1 (more specifically, delta~0.2).

adaptive/learner/averagelearner1D.py
AlvaroGI: Fixed in [this commit](https://github.com/python-adaptive/adaptive/pull/283/commits/4a1fa919b19ef378eae25b8620668b78f72d66a7)!

adaptive/learner/averagelearner1D.py
AlvaroGI: The current implementation of `tell_many()` only works if the elements of `ys` are lists. Is that a problem?

adaptive/learner/average_learner1D.py
```diff
@@ -0,0 +1,434 @@
...
+
+                    super()._update_scale(x, y_avg)
+                    self._update_losses_resampling(x, real=True)
+                    if self._scale[1] > self._recompute_losses_factor * self._oldscale[1]:
+                        for interval in reversed(self.losses):
+                            self._update_interpolated_loss_in_interval(*interval)
+                        self._oldscale = deepcopy(self._scale)
+
+        # If data is given:
+        else:
```
basnijholt, line 372: @AlvaroGI, why is this functionality necessary BTW?

AlvaroGI: This functionally allows to tell the learner about data that should be considered as "reliable", in the sense that it will not be resampled again. Theoretically, this data would correspond to the actual value of the function without noise, that is why there is no need to resample those points.
This functionality is not strictly necessary, but I was thinking in the following use-case scenario. Assume that you had some reliable previous data with low enough error bars that do not need resampling. If you want to add this data to the learner but you don't have access to all the samples anymore (only to the average value at each point), you can use this functionality, and the adaptive algorithm will take them into consideration when choosing a new point but will not resample them. I agree that this is not a common scenario but it may be useful for someone.

adaptive/learner/averagelearner1D.py
AlvaroGI: Fixed [here](https://github.com/python-adaptive/adaptive/pull/283/commits/f27cb09b862f9ffcec3b011d95122e0348f94843).

adaptive/learner/average_learner1D.py
basnijholt: I imagine that 99% of the people will exclusively use the `AverageLearner1D` with a `adaptive.Runner`. So in case this isn't really called, I vote for removing this option.

Maybe @akhmerov or @jbweston can chip in?

basnijholt: In any case, using `tell_many` to indicate that a point shouldn't be sampled anymore isn't a good API. If that method desired, there should probably be another method doing that.

adaptive/learner/average_learner1D.py
```diff
@@ -0,0 +1,435 @@
...
+
+        Examples
+        ----------
+        If xs = [0,1] and ys = [[1.2,1.4],[2]], the y-values will be included in
+        self._data_samples.
+        If xs = [0,1] and ys = [1.2,2], the y-values will be directly included
+        in self.data, and the previous self.data[0] and self.data[1] will be erased.
+        """
+        # Check that all x are within the bounds
+        assert np.prod([x>=self.bounds[0] and x<=self.bounds[1] for x in xs]), 'x value out of bounds: remove x or enlarge the bounds of the learner'
```
akhmerov, line 339: This may not be an `assert`: assertions are for internal sanity checks. Instead we should `raise ValueError` here.

adaptive/learner/average_learner1D.py
akhmerov: I agree with @basnijholt: this feature changes what the function means, and we shouldn't do this.

We may separately consider if driving the learner like this is a relevant feature to already implement, but it certainly shouldn't be here.

adaptive/learner/average_learner1D.py
```diff
@@ -0,0 +1,435 @@
...
+                    super()._update_scale(x, y_avg)
+                    self._update_losses_resampling(x, real=True)
+                    if self._scale[1] > self._recompute_losses_factor * self._oldscale[1]:
+                        for interval in reversed(self.losses):
+                            self._update_interpolated_loss_in_interval(*interval)
+                        self._oldscale = deepcopy(self._scale)
+
+        # If data is given:
+        else:
+            security_question = input('Function values given as scalars instead of lists. This will potentially overwrite the learner data. Continue? (y/n)')
```
akhmerov, line 373: Learners are meant to be used by other code: observe that `input` doesn't appear anywhere in adaptive library.

adaptive/learner/average_learner1D.py
```diff
@@ -0,0 +1,435 @@
...
+        t_student = scipy.stats.t.ppf(1 - self.alpha, df=n - 1)
+        return t_student * (variance_in_mean / n) ** 0.5
+
+    def tell_many(self, xs, ys):
+        """Tell the learner about some values.
+
+        Parameters
+        ----------
+        xs : Iterable of values from the function domain
+        ys : Iterable of lists of values (several data samples)
```
akhmerov, line 324: I think this breaks the API: runner isn't supposed to sort the points into lists with the same x-coordinate.

`tell` and `tell_many` must satisfy the relationship provided by the [BaseLearner](https://github.com/python-adaptive/adaptive/blob/master/adaptive/learner/base_learner.py#L87-L106). The reason why you may implement both is exclusively computational efficiency.

adaptive/learner/average_learner1D.py
```diff
@@ -0,0 +1,435 @@
...
+        is 1-2*alpha. We recommend to keep alpha=0.005.
+    neighbor_sampling : float (0 < neighbor_sampling <= 1)
+        Each new point is initially sampled at least a (neighbor_sampling*100)%
+        of the average number of samples of its neighbors.
+    min_samples : int (min_samples > 0)
+        Minimum number of samples at each point x. Each new point is initially
+        sampled at least min_samples times.
+    max_samples : int (min_samples < max_samples)
+        Maximum number of samples at each point x.
+    min_Delta_g : float (min_Delta_g >= 0)
```
basnijholt, line 34: Why is this parameter called this way? It's not a very obvious name (from reading this doc-string at least).

AlvaroGI: tl;dr: I agree, I changed the variable name and the doc-string [here](https://github.com/python-adaptive/adaptive/pull/283/commits/7550d1c64ae1d6d44c41b1d5774b5fba28c2d6cb). Do you think it is more clear now?

This variable was called min_Delta_g due to the original mathematical formulation of the problem. The true value of the function was called f(x), while the function with noise was g(x). The confidence intervals we build (the error bars) are for the mean value of g(x). Hence, I denoted the size of these intervals Δg. See [this notebook](https://nbviewer.jupyter.org/github/AlvaroGI/adaptive/blob/AverageLearner2D/test_strategy_5_new.ipynb) for a longer explanation with formulas.

adaptive/learner/average_learner1D.py
AlvaroGI: Makes sense! I removed this functionality in the [last implementation](https://github.com/python-adaptive/adaptive/pull/283/commits/a8c86ccae75e0f5d4da721bd1792f8c9a10c95e8) of `tell_many()` .

adaptive/learner/average_learner1D.py
AlvaroGI: Fixed [here](https://github.com/python-adaptive/adaptive/pull/283/commits/a8c86ccae75e0f5d4da721bd1792f8c9a10c95e8)!

adaptive/learner/average_learner1D.py
AlvaroGI: Makes sense, fixed [here](https://github.com/python-adaptive/adaptive/pull/283/commits/a8c86ccae75e0f5d4da721bd1792f8c9a10c95e8).

adaptive/learner/average_learner1D.py
AlvaroGI: I redesigned the `tell_many()` in the [last commit](https://github.com/python-adaptive/adaptive/pull/283/commits/a8c86ccae75e0f5d4da721bd1792f8c9a10c95e8), so that it matches the definition from the BaseLearner. I think now it works as it is supposed to.

adaptive/learner/average_learner1D.py
basnijholt: In other learners one is able to have numbers outside of the domain, the only thing that should happen is that it doesn't suggest points outside of the domain.

adaptive/learner/average_learner1D.py
basnijholt: Looks good! Thanks for the explanation.

basnijholt:
I have just rebased on top of master and rewritten history to remove the committed notebook outputs.

basnijholt:
I have added the `AverageLearner1D` to the tests but they are not passing yet.

One thing I had to do is to add `_get_data` and `_set_data` methods, however, when loaded the losses are different, so perhaps it is not yet correct. I will investigate further at a later point in time.

adaptive/learner/averagelearner1D.py
basnijholt: Seems fixed.

adaptive/tests/test_average_learner1d.py
```diff
@@ -0,0 +1,48 @@
...
+    assert len(learner._data_samples) == len(control._data_samples)
+    assert learner._data_samples.keys() == control._data_samples.keys()
+
+    for k, v1 in learner._data_samples.items():
+        v2 = control._data_samples[k]
+        assert len(v1) == len(v2)
+        np.testing.assert_almost_equal(np.sort(v1), np.sort(v2))
+
+    assert learner._bbox[0] == control._bbox[0]
+    assert learner._bbox[1] == control._bbox[1]
```
basnijholt, line 43: @AlvaroGI, @akhmerov, by adding this test I found a potential bug in the algorithm because it fails at this point.

I believe that the correct behavior for setting the bounding box in "y" is that it's set by the min and max average of the data.
This correctly happens in `tell_many_at_point` in `super()._update_scale(x, y_avg)`, however, when adding points one by one using `learner.tell` (starting with points far away from the mean) the bounding box will be defined by these points far away from the mean.

We can solve this by defining the bounding box:
- by its extremes `min([min(x) for x in learner._data_samples.values()]), max([max(x) for x in learner._data_samples.values()])` always (simplest solution but perhaps less correct)
- define the bounding box by the means of the data at each point, this means a lot of bookkeeping and perhaps (a lot?) extra computions

AlvaroGI: Hi Bas,
Sorry for the delay in my answer, I needed to take some time to think about this.

__Some findings after looking at this issue:__
 1. When the samples are added by running the learner, the maximum and minimum values of `learner._bbox` indeed correspond to the maximum and minimum values of `y` over all samples ("undesired" behavior), as you said. This is because the old `learner._bbox` values are compared to each new sample.
 2. However, when using `tell_many_at_point()`, I found that the `learner._bbox` values do *not* necessarily correspond to the maximum and minimum values of the data (data = average over all samples at each point).  In this case, old `learner._bbox` values are compared to the average of all the many samples that are added through `tell_many_at_point()`. Therefore, if you add points to a new learner, this corresponds to the maximum and minimum value of the data ("desired" behavior); but if we run a learner for a while and then add new samples, the values of `learner._bbox` will be compared to the mean value of the new samples, and we can end up with a `learner._bbox` that does not correspond to the max/min of the data nor to the max/min of the samples.
 3. Another bug I found in `tell_many_at_point()` is that the new value of the data at each point adopted the value of the mean of the new data samples, instead of the mean over all data samples (new and old) at that same point. I have already fixed this in my last commit.

I agree that the correct behavior should be the one you explained. However, I don't think `learner._bbox` has a dramatic impact on the performance of the algorithm (it is only used to compute `learner._scale`, which is mainly used to provide a finite loss to intervals with infinite loss, is that right?), so any of both behaviors could be fine in practice.
Therefore, I would say we should go for your first solution, which I think requires far less computations. To do that, I think we just need to bugfix `tell_many_at_point()` by replacing line 397 in `average_learner1D.py`:
```
super()._update_scale(x, self.data[x])
```
by the following two lines:
```
super()._update_scale(x, min(self._data_samples[x])
super()._update_scale(x, max(self._data_samples[x])
```

What do you think?

Issue #299, opened 2021-02-24
Runner fails in the notebook


ncuxomun:
Hi.

Thank you very much for sharing your project! 

I have attempted to run some tutorials on my machine(s), OS: Windows 10 | Python 3.8.5, but the runner fails to execute. For example:
![image](https://user-images.githubusercontent.com/54369128/109054190-71b62f80-7692-11eb-90fd-55ae9bc3d2be.png).

Please note that when I tried running the examples on Binder, everything worked smoothly. Can you please help me resolve the issue because I would like to run a set of my experiments?

Thank you very much for your attention!

basnijholt:
Could you try runner.task.print_stack() to see the error.

basnijholt:
Also see https://adaptive.readthedocs.io/en/latest/tutorial/tutorial.advanced-topics.html#debugging-problems

ncuxomun:
> Could you try runner.task.print_stack() to see the error.

Hi, Bas. May I call you by your name?
Thanks for responding. Here is the error:
![image](https://user-images.githubusercontent.com/54369128/109063512-125e1c80-769e-11eb-86ad-c6fbb82dcdfe.png)

basnijholt:
I think you are just missing the relevant bit of the error message here.

ncuxomun:
I screen'ed entire message in the notebook. How can I find the relevant bit?

basnijholt:
The part that comes after the raise RuntimeError on the last line.

ncuxomun:
I ran the script version, and the error message is:
![image](https://user-images.githubusercontent.com/54369128/109064430-4259ef80-769f-11eb-9333-352e7cdbbb6b.png)

ncuxomun:
![image](https://user-images.githubusercontent.com/54369128/109064806-b3010c00-769f-11eb-8a6f-f5531a286bb3.png)

basnijholt:
Could you try
```
pip install loky
```
and try again?

ncuxomun:
Hello!
Thank you so much for the suggestion - it worked! But I could not plot the figure as shown below:
![image](https://user-images.githubusercontent.com/54369128/109363522-a1f00080-7841-11eb-8b9e-692e7287aab5.png)
I did not observe the plotting issue when I was facing the previous issue.

basnijholt:
What version holoviews do you have installed?

ncuxomun:
[...]

Issue #292, opened 2020-09-02
can't pickle lru_cache function with loky


basnijholt:
The following fails:
```python
from functools import lru_cache
import adaptive
adaptive.notebook_extension()

@lru_cache
def g(x):
    return x

def f(x):
    return g(x)

learner = adaptive.SequenceLearner(f, range(2))
runner = adaptive.Runner(learner, adaptive.SequenceLearner.done)

runner.live_info()
```

Related to loky issue: https://github.com/joblib/loky/issues/268

This worked fine with the `concurrent.futures.ProcessPoolExecutor`.

basnijholt:
This is because of https://github.com/cloudpipe/cloudpickle/issues/178
and can be reproduced with
```python
import cloudpickle
from functools import lru_cache

@lru_cache
def g(x):
    return x

dump = cloudpickle.dumps(g)
del g
g = cloudpickle.loads(dump)
g(1)
```

akhmerov:
What's the use case for lru_cache'd functions?

basnijholt:
This occurs in some simulation software I use.

For example
```
@lru_cache
def make_kwant_syst():
    ...
    return syst

def f(x):
    syst = make_kwant_syst()
    return conductance(x, syst)
```

akhmerov:
I see. In the meantime you could hack around it by making `syst` global:

```python
def make_kwant_syst():
    try:
        return syst
    except NameError:
        global syst = ...
        return syst
```

basnijholt:
Also, it seems like this is blocked until at least the release of Python 3.9: https://github.com/cloudpipe/cloudpickle/pull/309#issuecomment-545472136.

Other alternatives include using your own memorization decorator:
```
def memoize(f):
    memo = {}
    def helper(x):
        if x not in memo:            
            memo[x] = f(x)
        return memo[x]
    return helper
```
or using the `concurrent.futures.ProcessPoolExecutor`.

akhmerov:
Should we then close this as an upstream bug? It seems like it will require no action either way.

basnijholt:
I think closing it will suggest that it's fixed. I rather leave it open with the "Blocked" label attached.

akhmerov:
But it's just not our bug, that is the reason to close it.

basnijholt:
Closing because it is an upstream issue
- https://github.com/joblib/loky/issues/268
- https://github.com/cloudpipe/cloudpickle/issues/178

PR #300, opened 2021-03-05
make loky a default on Windows and MacOS but not on Linux

basnijholt:
## Description

Make loky a default on Windows and MacOS but not on Linux.

I am doing this because several people have indicated that they are experiencing issues with Loky on a Linux cluster.
On Windows and MacOS loky is preferred because `ProcessPoolExecutor` won't work in interactive environments.

Closes https://github.com/python-adaptive/adaptive/issues/301

See also
- https://github.com/python-adaptive/adaptive/issues/292
- https://github.com/python-adaptive/adaptive/issues/299

## Checklist

- [ ] `pytest` passed

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -5,10 +5,13 @@ import functools
 import inspect
 import itertools
 import pickle
+import platform
 import time
 import traceback
 import warnings
 from contextlib import suppress
+
+import loky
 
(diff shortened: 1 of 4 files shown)


akhmerov:
Sounds good to me.

PR #311, opened 2021-03-19
lazily evaluate the integrator coefficients

basnijholt:
## Description

Currently, the having the following in a script:
```
import adaptive
import gpytorch
```
breaks any Python process because of https://github.com/pytorch/pytorch/issues/54063.

This is not a proper fix but avoiding the problem. 

Howver, rather than doing any math on import, I think it is a better style to only evaluate the coefficients when required.

## Checklist

- [x] `pytest` passed

## Type of change

*Check relevant option(s).*

- [x] (Code) style fix or documentation update

### DIFF
--- a/adaptive/learner/integrator_coeffs.py
+++ b/adaptive/learner/integrator_coeffs.py
@@ -141,39 +141,105 @@ def calc_V(x, n):
     return np.array(V).T
 
 
-eps = np.spacing(1)
+class Coefficients:
+    def __init__(self) -> None:
+        self.is_set = False
+        # The nodes
+        self.ns = (5, 9, 17, 33)
+        self.eps = np.spacing(1)
 
-# the nodes and Newton polynomials
-ns = (5, 9, 17, 33)
-xi = [-np.cos(np.linspace(0, np.pi, n)) for n in ns]
+        # If the relative difference between two consecutive approximations is
+        # lower than this value, the error estimate is considered reliable.
+        # See section 6.2 of Pedro Gonnet's thesis.
+        self.hint = 0.1
 
-# Make `xi` perfectly anti-symmetric, important for splitting the intervals
-xi = [(row - row[::-1]) / 2 for row in xi]
+        # Smallest acceptable relative difference of points in a rule.  This was chosen
+        # such that no artifacts are apparent in plots of (i, log(a_i)), where a_i is
+        # the sequence of estimates of the integral value of an interval and all its
+        # ancestors..
+        self.min_sep = 16 * self.eps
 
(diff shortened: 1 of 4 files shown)


akhmerov:
Introducing the class and requiring that consumers instantiate it makes the code harder to use.

What do you think about using module-level `__getattr__` instead, see [here](https://stackoverflow.com/a/57263518/2217463)?

akhmerov:
Additionally I don't follow the reasoning why not doing linear algebra on import is better style.

- The new implementation is certainly harder to read, which is usually a sign of worse style.
- Also according to my rough benchmarks the import time isn't adversely affected.
- Since we are not mutating anything but the module namespace, there are no side-effects, which is the usual argument against doing stuff on import (pyplot I'm looking at you!)

Why do you claim it is better?

basnijholt:
> What do you think about using module-level __getattr__ instead, see here?

Two reasons why it might not be better.
1. It's Python ≥3.7 only and we require ≥3.6.
2. If we decide to cache the values (which we should) I think we need to evaluate all coefficients as many times as there are coefficients.

> Since we are not mutating anything but the module namespace, there are no side-effects, which is the usual argument against doing stuff on import (pyplot I'm looking at you!)

This is not true and the whole reason I would like to make this change. Calling a `scipy` function loads the BLAS/MKL implementation. In case one wants to import a package that was compiled against a different BLAS implementation, stuff might go wrong, as happens in https://github.com/pytorch/pytorch/issues/54063.

basnijholt:
@akhmerov, that's a really good suggestion!

I've implemented that and opened https://github.com/python-adaptive/adaptive/pull/312 to bump our requirement to Python≥3.7.

basnijholt:
I have rebased this on top of the latest master (which has the Python ≥ 3.7 requirement) so all tests should now pass.

adaptive/learner/integrator_coeffs.py
```diff
@@ -141,39 +142,48 @@ def calc_V(x, n):
     return np.array(V).T
 
 
-eps = np.spacing(1)
+@lru_cache(maxsize=None)
+def _coefficients():
```
akhmerov, line 146: One last request: let us provide a minimal docstring explaining the purpose of this function. Something like "Compute the coefficients on demand, in order to avoid doing linear algebra on import."

akhmerov: Aside of this everything looks good!

basnijholt: Great! I have just added that description.

PR #317, opened 2021-04-06
add to_numpy methods

basnijholt:
## Description

After having answered "How do I get the data" 1000 times and having written the code to convert `learner.data` to a numpy array many times as well, I decided to just add it in Adaptive.

I have implemented it such that a single array is returned with the points and values concatenated. An alternative is to return the points and values as separate arrays.

## Checklist

- [x] `pytest` passed

## Type of change

*Check relevant option(s).*


jbweston (approved):
LGTM. Should we also add this to the documentation?

basnijholt:
Should be in the API docs :-)

Thanks for merging!

PR #320, opened 2021-05-26
call self._process_futures on canceled futures when BlockingRunner is done

### LINKED ISSUE #319: Runners should tell learner about remaining points at end of run
Currently, `BlockingRunner` will [wait for all futures](https://github.com/python-adaptive/adaptive/blob/4f2200bbaa4e65459df4ce4196182ee0c15c4666/adaptive/runner.py#L414) that are running when the goal is reached, but then ignore their results. Looking at the code, `AsyncRunner` has the same behaviour.

This seems like a waste. I think it should just `self._process_futures(remaining)` after the wait.

basnijholt:
## Description

call self._process_futures on canceled futures when BlockingRunner is done.

thanks @jgukelberger for noticing this

Fixes #319

## Checklist

- [x] `pytest` passed

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -412,7 +412,10 @@ class BlockingRunner(BaseRunner):
             remaining = self._remove_unfinished()
             if remaining:
                 concurrent.wait(remaining)
-                self._process_futures(remaining)
+                # Some futures get their result set, despite being cancelled.
+                # see https://github.com/python-adaptive/adaptive/issues/319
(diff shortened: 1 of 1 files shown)


akhmerov (approved):

basnijholt:
It appears that in our tests, the futures of a `BlockingRunner` do get canceled as expected. So I added a check to filter out the futures that are done.

PR #340, opened 2021-12-13
Fix NaN issue for Learner1D R -> R^n

### LINKED ISSUE #339: Learner1D R -> R^n - Infinite loop
I'm running a series of tests and the following function (which is not defined at x=0) is giving me troubles, as the evaluation loop keeps going on forever:

```python
from adaptive import Learner1D
from adaptive.runner import simple
import numpy as np

def func(x):
    w = -1j * np.exp(1j * np.pi * x) / x + 1j * np.exp(-1j * np.pi * x) / x
    return np.real(w), np.imag(w)

learner = Learner1D(func, bounds=(-10.0, 10.0))
simple(learner, lambda l: l.loss() < 0.01)
learner.to_numpy()
```

By inspecting the `default_loss` function, this is what is going on:

```python
def default_loss(xs, ys):
    print("xs", xs)
    print("ys", ys)
    
    dx = xs[1] - xs[0]
    if isinstance(ys[0], collections.abc.Iterable):
        dy_vec = np.array([abs(a - b) for a, b in zip(*ys)])
        return np.hypot(dx, dy_vec).max()
    else:
        dy = ys[1] - ys[0]
        return np.hypot(dx, dy)

learner = Learner1D(func, bounds=(-10.0, 10.0), loss_per_interval=default_loss)
simple(learner, lambda l: l.loss() < 0.01)
learner.to_numpy()

# OUTPUT:
# xs (-0.5, 0.5)
# ys (array([-2.4492936e-16,  0.0000000e+00]), array([-2.4492936e-16,  0.0000000e+00]))
# xs (-0.5, 0.0)
# ys (array([nan, nan]), array([nan, nan]))
# xs (0.0, 0.5)
# ys (array([nan, nan]), array([nan, nan]))
# xs (-0.5, -0.25)
# ys (array([nan, nan]), array([nan, nan]))
```

Seems like there is something wrong with the function's evaluation. For example `func(np.array([-0.5, -0.25]))` should compute `(array([4.        , 5.65685425]), array([0., 0.]))` whereas `ys = (array([nan, nan]), array([nan, nan]))`.

Davide-sd:
## Description

If a function evaluates over a point where it is not defined, NaN will be returned.
In the case of a function R -> R^n, it caused `_update_scale` to compute `self._scale[1] = nan`,
which in turns led to the evaluation of the function on all other points to be NaN, causing an
infinite loop, as reported in issue #339.

Using `np.nanmin` and `np.nanmax` in place of `np.min` and `np.max` seems to solve the issue.

Fixes  #339

## Checklist

- [ ] `pytest` passed

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/learner/learner1D.py
+++ b/adaptive/learner/learner1D.py
@@ -464,8 +464,8 @@ class Learner1D(BaseLearner):
         if y is not None:
             if self.vdim > 1:
                 try:
-                    y_min = np.min([self._bbox[1][0], y], axis=0)
-                    y_max = np.max([self._bbox[1][1], y], axis=0)
+                    y_min = np.nanmin([self._bbox[1][0], y], axis=0)
+                    y_max = np.nanmax([self._bbox[1][1], y], axis=0)
                 except ValueError:
                     # Happens when `_bbox[1]` is a float and `y` a vector.
                     y_min = y_max = y


basnijholt:
Thanks a lot for the fix @Davide-sd!

basnijholt (approved):
LGTM!

Davide-sd:
Thanks @basnijholt .

I've run all tests twice on my local machine. In the first run everything appeared to be correct. On the second run the `adaptive/tests/test_runner.py::test_ipyparallel_executor` failed. Not sure why...

Issue #352, opened 2022-07-09
Stop using atomic writes


akhmerov:
[atomicwrites is now deprecated](https://github.com/untitaker/python-atomicwrites#unmaintained) and recommends using `os.replace` instead. pytest used pytest-dev/pytest#10115 to make a replacement. While I don't expect that things would break on a short notice, we should consider stopping to rely on atomicwrites.

basnijholt:
Well caught. I'll fix this on a slow day.

basnijholt:
Fixed by #353 🎉

PR #353, opened 2022-07-18
Replacing atomicwrites with os.write 

juandaanieel:
Solves #352


akhmerov:
I don't think this solves the problem of atomic writes, or at least I don't understand why it would. Please compare with the pytest solution that I linked in the issue.

juandaanieel:
Please check the new version.
It works as in pytest with a temporary file.

basnijholt (approved):
Thanks a lot, looks good!

PR #358, opened 2022-09-06
Add getting learner's data as pandas.DataFrame; add learner.to_dataframe method

basnijholt:
## Description

This code in this PR allows exporting a learner to a `pandas.DataFrame`. This will be useful when running many learners (e.g, in a `BalancingLearner`) and then being able to just concatenate the `DataFrame`s to get the combined result.

Currently saving the learner creates a `pickle` file and while this is fine for immediate use, for long term storage this is a bad practice.

I chose to use `pandas` over `xarray` because we are dealing with tabular data instead of ND dense or sparse arrays.

Ping @CameronKing

Example:
image

## Checklist

- [ ] `pytest` passed

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/tests/test_learners.py
+++ b/adaptive/tests/test_learners.py
@@ -714,7 +714,7 @@ def test_to_dataframe(learner_type, f, learner_kwargs):
     learner = learner_type(generate_random_parametrization(f), **learner_kwargs)
 
     # Test empty dataframe
-    df = learner.to_dataframe()
+    df = learner.to_dataframe(**kw)
     assert len(df) == 0
 
     # Run the learner
--- a/adaptive/tests/test_learners.py
+++ b/adaptive/tests/test_learners.py
@@ -783,3 +783,13 @@ def test_to_dataframe(learner_type, f, learner_kwargs):
         assert len(df) == data_saver.nsamples
     else:
         assert len(df) == data_saver.npoints
+
+    # Test loading from a DataFrame into a new DataSaver
+    learner2 = learner_type(learner.function, **learner_kwargs)
+    data_saver2 = DataSaver(learner2, operator.itemgetter("result"))
+    data_saver2.load_dataframe(df, **kw)
+    assert data_saver2.extra_data.keys() == data_saver.extra_data.keys()
+    assert all(
+        data_saver2.extra_data[k] == data_saver.extra_data[k]
+        for k in data_saver.extra_data.keys()
+    )
--- a/adaptive/tests/test_learners.py
+++ b/adaptive/tests/test_learners.py
@@ -736,12 +736,19 @@ def test_to_dataframe(learner_type, f, learner_kwargs):
         learner_type(generate_random_parametrization(f), **learner_kwargs)
         for _ in range(2)
     ]
-    learner = BalancingLearner(learners)
-    simple_run(learner, 100)
-    df = learner.to_dataframe(**kw)
-    assert isinstance(df, pandas.DataFrame)
+    bal_learner = BalancingLearner(learners)
+    simple_run(bal_learner, 100)
+    df_bal = bal_learner.to_dataframe(**kw)
+    assert isinstance(df_bal, pandas.DataFrame)
 
     if learner_type is not AverageLearner1D:
-        assert len(df) == learner.npoints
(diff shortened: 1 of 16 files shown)


akhmerov:
What problem does this PR solve? I believe a pandas DataFrame isn't a go-to choice for numerical data; is there a use case which is sufficiently broadly relevant?

akhmerov:
OK, I see: the entries indeed have different meanings and possibly dimensions (x, y, value). For that reason pandas seems appropriate.

adaptive/utils.py
```diff
@@ -98,3 +99,43 @@ def __call__(self, *args, **kwargs):
...
+            " it is used to distinguish between function and learner parameters."
+        )
+    kwargs = {}
+    for col in df.columns:
+        if col.startswith(function_prefix):
+            k = col.split(function_prefix, 1)[1]
+            vs = df[col]
+            v, *rest = vs.unique()
+            if rest:
+                raise ValueError(f"The column '{col}' can only have one value.")
```
jbweston, line 137: Just realized that this means that any default values attached to the learner's function will be repeated in the dataframe for every point the learner evaluates!

I guess the reason for this is to facilitate combining results from learners where the default values are different, but this clause shows that `partial_function_from_dataframe` will fail in that case.

Is it possible to attach "metadata" to an entire dataframe (xarrays have this) to avoid the redundancy?

adaptive/tests/test_learners.py
```diff
@@ -694,3 +695,60 @@ def test_learner_subdomain(learner_type, f, learner_kwargs):
...
+
+    # Test this for a learner in a BalancingLearner
+    learners = [
+        learner_type(generate_random_parametrization(f), **learner_kwargs)
+        for _ in range(2)
+    ]
+    bal_learner = BalancingLearner(learners)
+    simple_run(bal_learner, 100)
+    df_bal = bal_learner.to_dataframe(**kw)
+    assert isinstance(df_bal, pandas.DataFrame)
```
jbweston, line 742: What happens if you mix learner types in a balancing learner?

jbweston (approved):

adaptive/tests/test_learners.py
basnijholt: We do not support that:
https://github.com/python-adaptive/adaptive/blob/c59405ac12f2422c0e6b014ab76fd7683984556f/adaptive/learner/balancing_learner.py#L84-L87

adaptive/utils.py
```diff
@@ -98,3 +99,43 @@ def __call__(self, *args, **kwargs):
                     msg = f"The attribute '{name}' should be of type {type_}, not {type(x)}."
                     raise TypeError(msg)
         return obj
+
+
+def _default_parameters(
+    function, function_prefix: str = "function.", start_index: int = 1
```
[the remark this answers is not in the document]
basnijholt: I just had this for generality but I am fine with removing it.

Note that this is a private function though.

adaptive/utils.py
basnijholt: Using constant columns is a standard practice.

It is correct that using the `partial_function_from_dataframe` directly won't work and I would also expect it to not work.

The `BalancingLearner` does a `df.groupby` operation and then uses `partial_function_from_dataframe`.

Since pandas 1.0 there is a metadata attribute `df.attrs` however, they won't combine easily. I think one of the biggest advantages to this new "learners to dataframes" feature is precisely being able to combine learners easily, so I prefer the current implementation.

basnijholt: We can use the `category` datatype:
image

akhmerov: Not saving parameters that the learner doesn't vary is another option to consider. It makes the code simpler and avoids handling special cases.

basnijholt: I think we should keep it in, there aren't that many special cases and the code is quite simple.

adaptive/utils.py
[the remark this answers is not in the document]
jbweston: Yeah take it out; we can put it back in if we need it later.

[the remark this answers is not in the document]
basnijholt: Good, done! 👍

akhmerov (approved):

PR #381, opened 2022-10-14
SequenceLearner: Use a dictionary instead of a SortedSet for _to_do_indices

basnijholt:
We only ever create the set and then pop elements from it (except in `remove_unfinished`). Since dictionaries preserve insertion order now, we can benefit from increased performance.

```python
%time x = dict.fromkeys(range(int(20e6)))
```
takes 1.6s

and
```python
from sortedcontainers import SortedSet
%time x = SortedSet(range(int(20e6)))
```
takes 5.4s

## Description

Please include a summary of the change and which (if so) issue is fixed.

Fixes #(ISSUE_NUMBER_HERE)

## Checklist

- [ ] `pytest` passed

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/learner/sequence_learner.py
+++ b/adaptive/learner/sequence_learner.py
@@ -5,7 +5,7 @@ from numbers import Integral as Int
 from typing import Any, Tuple
 
 import cloudpickle
-from sortedcontainers import SortedDict, SortedSet
+from sortedcontainers import SortedDict
 
 from adaptive.learner.base_learner import BaseLearner
 from adaptive.utils import assign_defaults, partial_function_from_dataframe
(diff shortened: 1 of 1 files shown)


basnijholt:
I did some benchmarking:
`test.py`:
```
from adaptive import SequenceLearner, runner
from toolz import identity
from time import time
seq = range(int(1e6))
t = time()
learner = SequenceLearner(identity, seq)
runner.simple(learner, SequenceLearner.done)
print(time() - t)
```

```
basnijholt-macbook  ➜  adaptive git:(main) ✗  python test.py
0.29935383796691895
basnijholt-macbook  ➜  adaptive git:(main) ✗  gco dict-instead-of-sortedset
Switched to branch 'dict-instead-of-sortedset'
Your branch is up to date with 'origin/dict-instead-of-sortedset'.
basnijholt-macbook  ➜  adaptive git:(dict-instead-of-sortedset) ✗  python test.py
1.7948999404907227
```


So it turns out to be way slower. Closing this PR.

PR #382, opened 2022-10-14
Add loss_goal, npoints_goal, and an auto_goal function and use it in the runners

basnijholt:
## Description

Add more loss options to Runners to easily set common loss conditions.

For example:
```python
learner = Learner1D(...)
runner = Runner(learner, npoints_goal=100)
# or
runner = Runner(learner, loss_goal=0.01)
```

```python
learner = SequenceLearner(...)
runner = Runner(learner)  # stops when out of points
```

```python
learner = IntegratorLearner(..., rtol=0.01)
runner = simple(learner)  # stops when reached tol
```

```python
learner = BalancingLearner([list of IntegratorLearners])
runner = Runner(learner)  # stops when all reached tol
```

```python
learner = Learner1D(...)
runner = Runner(learner)  # never stops
```

```python
learner = SequenceLearner(...)
runner = Runner(learner, duration_goal=timedelta(hours=5))
```

Still one is also able to use `auto_goal`:
```python
from adaptive.runner import auto_goal
learner = Learner2D(...)
runner = Runner(learner, auto_goal(loss=0.01))
runner = Runner(learner, auto_goal(npoints=100))
runner = Runner(learner, auto_goal(timedelta=timedelta(hours=1))
```

## Checklist

- [x] `pytest` passed

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -13,6 +13,7 @@ import time
 import traceback
 import warnings
 from contextlib import suppress
+from datetime import datetime, timedelta
 from typing import TYPE_CHECKING, Any, Callable
 
 import loky
@@ -129,10 +130,11 @@ class BaseRunner(metaclass=abc.ABCMeta):
         shutdown_executor=False,
         retries=0,
         raise_if_retries_exceeded=True,
+        allow_running_forever=False,
     ):
 
         self.executor = _ensure_executor(executor)
-        self.goal = goal
+        self.goal = auto_goal(goal, learner, allow_running_forever)
 
         self._max_tasks = ntasks
 
@@ -396,6 +398,7 @@ class BlockingRunner(BaseRunner):
             shutdown_executor=shutdown_executor,
             retries=retries,
             raise_if_retries_exceeded=raise_if_retries_exceeded,
+            allow_running_forever=False,
         )
         self._run()
 
@@ -518,11 +521,6 @@ class AsyncRunner(BaseRunner):
         raise_if_retries_exceeded=True,
     ):
 
-        if goal is None:
-
-            def goal(_):
-                return False
-
         if (
             executor is None
             and _default_executor is concurrent.ProcessPoolExecutor
@@ -548,6 +546,7 @@ class AsyncRunner(BaseRunner):
             shutdown_executor=shutdown_executor,
             retries=retries,
             raise_if_retries_exceeded=raise_if_retries_exceeded,
+            allow_running_forever=True,
         )
         self.ioloop = ioloop or asyncio.get_event_loop()
         self.task = None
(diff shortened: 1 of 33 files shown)


adaptive/runner.py
```diff
@@ -861,3 +861,89 @@ def _get_ncores(ex):
...
+    Parameters
+    ----------
+    goal
+        The goal to extract. Can be a callable, an integer, a float, a datetime,
+        a timedelta or None.
+        If it is a callable, it is returned as is.
+        If it is an integer, the goal is reached after that many points have been
+        returned.
+        If it is a float, the goal is reached when the learner has reached a loss
+        less than that.
```
akhmerov, line 898: This part seems fragile: the distinction between a float an an int isn't big enough.

akhmerov (commented):
I like the idea of simplifying common use cases, and therefore I find the MR promising, however I think that the logic for interpreting the input is too obscure and would hurt readability. Instead we could introduce alternative arguments to the runner, e.g. `Runner(..., time=...)` etc.

akhmerov (changes requested):
I thought about it more, and I'm pretty sure that bundling all options into goal makes the code less readable and more error-prone. I think adding to the runner a handful of arguments that specify different termination conditions is the way to go.

There are examples where input interpretation depends on its shape, in particular a bunch of numpy API, but I don't know where as much of interpretation would depend on input types.

adaptive/runner.py
basnijholt: I just encountered this today https://scikit-learn.org/stable/modules/generated/sklearn.cluster.OPTICS.html#sklearn.cluster.OPTICS

image

akhmerov: Fair point, however in this example also the ranges do not overlap. Loss, on the other hand, can assume arbitrary values.

basnijholt:
@akhmerov, I have added `loss_goal` and `npoints_goal` arguments to all the runners we provide, it will cover most of the use cases. We're also using it everywhere in the tutorials and docs now.

For example:
```python
learner = Learner1D(...)
runner = Runner(learner, npoints_goal=100)
# or
runner = Runner(learner, loss_goal=100)
```
However, I hope you are willing to compromise because I would like to leave the `goal` where it can accept:
```python
GoalTypes = Union[Callable[[BaseLearner], bool], int, float, datetime, timedelta, None]
```
Some examples:
```python
learner = SequenceLearner(...)
runner = Runner(learner)  # stops when out of points
```

```python
learner = IntegratorLearner(..., rtol=0.01)
runner = Runner(learner)  # stops when reached tol
```

```python
learner = BalancingLearner([list of IntegratorLearners])
runner = Runner(learner)  # stops when all reached tol
```

```python
learner = SequenceLearner(...)
runner = Runner(learner, goal=timedelta(hours=5))
```

Still one is able to do (while also having the option of being more explicit)
```python
learner = Learner2D(...)
runner = Runner(learner, goal=0.01)
```

I believe the `auto_goal` function would be less useful without the `loss` and `npoints` options.

BTW [Adaptive Scheduler has a similar interface](https://github.com/basnijholt/adaptive-scheduler/blob/d7a66880c53346e6831178e902fba6025ec42df0/adaptive_scheduler/server_support.py#L853-L858). I discussed this interface with quite a few heavy Adaptive users, and all of them think this is a great convenience.

Perhaps @jbweston can chime in too; do you think this is an OK compromise?

basnijholt:
@akhmerov, I had a long discussion with @jbweston today and we settled on what I added in the description of this PR.

tl;dr, he agreed with you and now we are being explicit.

akhmerov:
Thanks! A naming question: do we want to keep `_goal` in the parameter names? `npoints` seems clear enough. On the other hand `timedelta` is a type, `duration` would be clearer, and it should probably accept float-like numbers of seconds too. Likewise `datetime` could be `end_time`. `auto_goal` seems :ok_hand:.

basnijholt:
I,
* renamed `timedelta_goal` to `duration_goal`,
* renamed `datetime_goal` to `end_time_goal`,
* allowed `duration_goal` to be a number, which indicates the number of seconds.

I think having the `_goal` explicit in the `Runner` aguments is clearer because it groups all of those arguments together.

docs/source/tutorial/tutorial.Learner1D.md
```diff
@@ -124,7 +124,7 @@ The `Learner1D` can be used for such functions:
 
 ```{code-cell} ipython3
 learner = adaptive.Learner1D(f_levels, bounds=(-1, 1))
-runner = adaptive.Runner(learner, goal=lambda l: l.loss() < 0.01)
+runner = adaptive.Runner(learner, loss_goal=0.01)  # continue until `learner.loss()<=1`
```
jbweston, line 127:
```suggestion
runner = adaptive.Runner(learner, loss_goal=0.01)  # continue until `learner.loss()<=0.01`
```

jbweston (approved):
Nice feature!

There are a few minor fixups that could be applied, but other thatn that this is good to merge.

akhmerov (approved):
All looks good, thanks. I still think that `_goal` is superfluous in some of the new arguments like `duration` or `end_time`, but I leave this for you to decide.

basnijholt:
Like I mentioned in https://github.com/python-adaptive/adaptive/pull/382#issuecomment-1324062053 I think keeping the `goal` in the parameter name makes it clear that these parameters are grouped (@jbweston agreed in an offline discussion):

- `goal`
- `loss_goal`
- `npoints_goal`
- `end_time_goal`
- `duration_goal`

And it will be clear that they have a different meaning from the other parameters of e.g., the `AsyncRunner`:

- `executor`
- `ntasks`
- `log`
- `shutdown_executor`
- `ioloop`
- `retries`
- `raise_if_retries_exceeded`

akhmerov:
I see, thanks for the explanation. I agree with the reasoning.

PR #440, opened 2023-07-17
Ensure periodic saving fires immediately after runner task is finished

### LINKED ISSUE #439: large delay when using start_periodic_saving
The current implementation of 'start_periodic_saving' does `await sleep(interval)`:

https://github.com/python-adaptive/adaptive/blob/88f02588cf66d0276b9b77da466aa71c7c1f1773/adaptive/runner.py#L839C1-L843C50

This means that there can be a large `interval`-length delay between the time the runner completes and the time that the last write is made by `start_periodic_saving`.

This is very confusing when you wait for the runner to finish with `await runner.task` and the results are not yet written.

Fix is to have the `await` await both the timeout and `runner.task` (remembering to shield `runner.task`)

jbweston:
## Description

This PR modifies the behavior of periodic saving to fire immediately after the runner completes.
Previously the periodic saving would wait 'interval' seconds before firing the final save event.

Fixes #439 .

## Checklist

- [x] `pytest` passed

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -839,7 +839,8 @@ class AsyncRunner(BaseRunner):
         async def _saver():
             while self.status() == "running":
                 method(self.learner)
-                await asyncio.sleep(interval)
+                # No asyncio.shield needed, as 'wait' does not cancel any tasks.
+                await asyncio.wait([self.task], timeout=interval)
             method(self.learner)  # one last time
 
         self.saving_task = self.ioloop.create_task(_saver())
--- a/docs/source/tutorial/tutorial.IntegratorLearner.md
+++ b/docs/source/tutorial/tutorial.IntegratorLearner.md
@@ -4,7 +4,7 @@ jupytext:
     extension: .md
     format_name: myst
     format_version: 0.13
-    jupytext_version: 1.14.5
+    jupytext_version: 1.14.7
 kernelspec:
   display_name: python3
   name: python3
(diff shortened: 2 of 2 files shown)


akhmerov (approved):
Cool implementation!

What will this do if the `runner.task` is canceled? Do I understand correctly that it'll still save one last time?

adaptive/runner.py
```diff
@@ -839,7 +839,8 @@ def default_save(learner):
         async def _saver():
             while self.status() == "running":
                 method(self.learner)
-                await asyncio.sleep(interval)
+                # No asyncio.shield needed, as 'wait' does not cancel any tasks.
```
basnijholt, line 842: LGTM 👍 

Don't understand what this comment is adding though?

basnijholt (approved):

adaptive/runner.py
jbweston: There is another function named `wait_for`, which waits for the first task to complete and then cancels all the other ones.
If you want it not to cancel the other tasks you need to wrap them in `asyncio.shield`.

My comment is meant to say that this (the not cancelling) is indeed the behavior we want, and that `shield` is not needed in order to achieve this behavior.

jbweston:
> Cool implementation!
> 
> What will this do if the `runner.task` is canceled? Do I understand correctly that it'll still save one last time?

It will indeed save one last time.

I believe this to be the semantics we want; we want all the results that the learner has at a particular point in time to be reflected in the saved results, regardless of whether the runner was cancelled, or stopped of its own accord.

adaptive/runner.py
basnijholt: OK great. Thanks for clarifying! :)

PR #477, opened 2025-05-13
Follow SPEC 0 and drop support for Python 3.9 and 3.10

basnijholt:
According to https://scientific-python.org/specs/spec-0000/ we should even drop support for 3.10 already.


akhmerov:
I'm happy with fully following spec-0.

basnijholt:
@akhmerov, just did that 😄

PR #484, opened 2025-08-20
Add customizable number of points in simple runner.

krokosik:
## Description

My use case has a function that is already fast due to parallel execution. As such, not only does it evaluate quicker than 50ms, but also freezes my PC when run with parallel runner. As such, I found the `simple` runner very useful, but it also introduces overhead by running the interpolants on ever new point. Increasing the `ask` count to 16 (my number of cores to mimic the parallel version), diminished that to overhead to almost nothing, but I would prefer to not have to monkey patch this function :)

## Checklist

- [x] `pytest` passed

## Type of change

*Check relevant option(s).*

### DIFF
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -911,7 +911,7 @@ def simple(
     npoints_goal: int | None = None,
     end_time_goal: datetime | None = None,
     duration_goal: timedelta | int | float | None = None,
-    points_per_ask: int | None = 1,
+    points_per_ask: int = 1,
 ):
     """Run the learner until the goal is reached.
 
--- a/adaptive/runner.py
+++ b/adaptive/runner.py
@@ -911,11 +911,12 @@ def simple(
     npoints_goal: int | None = None,
     end_time_goal: datetime | None = None,
     duration_goal: timedelta | int | float | None = None,
+    points_per_ask: int | None = 1,
 ):
     """Run the learner until the goal is reached.
 
-    Requests a single point from the learner, evaluates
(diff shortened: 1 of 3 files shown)


adaptive/runner.py
```diff
@@ -911,11 +911,12 @@ def simple(
     npoints_goal: int | None = None,
     end_time_goal: datetime | None = None,
     duration_goal: timedelta | int | float | None = None,
+    points_per_ask: int | None = 1,
```
akhmerov:
```suggestion
    points_per_ask: int = 1,
```

akhmerov (changes requested):
This is nice and simple, looks helpful too! Please add a test to confirm that `points_per_ask > 1` works, and I think this PR is good to go!

adaptive/runner.py
krokosik: This would make the type checker yell at the end user for not providing this arguments.

krokosik:
While I understand the need for full test coverage, this is a bit problematic here. The `simple` runner is basically a blocking while loop and there is no way to check the `pending_points` after `ask` has been called.

I guess we could mock this method, but I'm not that experienced with `pytest`, I guess it would involve adding the `pytest-mock` plugin, but do we want to go through that for such a simple functionality that does not introduce any changes for existing code?

basnijholt (approved):
Made the change @akhmerov suggested (which I fully agree with). @krokosik I don't see how this is a problem with type checkers?

Also added a test.

akhmerov (approved):

PR #485, opened 2025-08-21
WIP: Add Rust implementation of triangulation module using PyO3

basnijholt:
This is just an experiment and 100% vibe coded... I want to see whether we can get a significant speedup by implementing the LearnerND in Rust.


akhmerov:
What were your findings?

basnijholt:
That in (N>2)D SciPy is actually really fast, it was like >100x faster then the Rust code but in 2D Rust was 5-30x faster. Didn't do a lot of validation of whether it was doing the optimal thing though.

akhmerov:
I thought scipy triangulation isn't designed for incremental updates; did you test on adding points one at a time or all at once?

PR #486, opened 2025-10-01
Canonicalise LearnerND hull seeding without duplicate collisions

### LINKED ISSUE #470: Question: Running a learning for additional points after some amount of points was computed
Hi

My question is the following:

how one could run a learner again after sampling a certain amount of points.

I have tried following:

run LearnerND for sampling 2000 points.

Saved them to the dataframe

Started new run.

Created new instance of LeanerND with same initial parameters and function.

Loaded points generated from old session using 
```
learner.load_dataframe(initial_df,
                           with_default_function_args=False,
                           point_names=("x", "y"),
                           value_name="value"

)
```

and tried running

```
    points = len(initial_df)+110
    runner=BlockingRunner(learner, npoints_goal=points)
    df = learner.to_dataframe(point_names=("x", "y"))
```
However, I get the exception:
```
raise ValueError("Point already in triangulation.")
```
from add_point (line 629, in add_point, called from line 605, in _try_adding_pending_point_to_simplex)

basnijholt:
Closes #470


adaptive/tests/unit/test_learnernd_integration.py
```diff
@@ -53,3 +58,29 @@ def test_learnerND_log_works():
     learner.ask(2)
     # At this point, there should! be one simplex in the triangulation,
     # furthermore the last two points that were asked should be in this simplex
+
+
+@pytest.mark.skipif(not with_pandas, reason="pandas is not installed")
+def test_learnerND_resume_after_loading_dataframe_convex_hull():
```
basnijholt: before merging let's delete the data again

